PR #29, opened 2026-06-14
refactor: generate OpenAPI spec from utoipa decorators

tupe12334:
## Summary

- Delete `src/build/openapi.rs` (323-line hand-authored JSON literal) — the source of drift between docs and code
- Add `src/openapi.rs` with `ApiDoc` (`#[derive(utoipa::OpenApi)]`) that collects all paths and schemas from the existing `#[utoipa::path]` / `#[utoipa::ToSchema]` decorators
- Extract `index`, `health`, `echo` from inline closures to named `pub` handlers with `#[utoipa::path]` decorators; add `HealthResponse`, `EchoRequest`, `EchoResponse` with `ToSchema`
- Add `cron_jobs::replace` (thin PUT alias over `update`) so PUT and PATCH are both documented separately
- Move `uptime_start` into `AppState` so the extracted `health` handler can read it via `State<AppState>`
- Write `apis/openapi.json` at server startup (`run_with_listener_until`) instead of at build time — the committed file now reflects the decorators automatically whenever the server runs
- `SwaggerUi` at `/docs` points to `crate::openapi::ApiDoc` instead of a local private struct
- Drop the source-file `rerun-if-changed` entries from `build.rs` that only existed for the old hardcoded generation

## Test plan

- [ ] `cargo build` passes (no errors)
- [ ] `cargo test` — all 155 tests pass
- [ ] `apis/openapi.json` regenerates correctly after running the server or test suite
- [ ] `/docs` Swagger UI reflects all endpoints including `PUT /cron-jobs/{id}`, `GET /`, `GET /health`, `POST /echo`


tupe12334:
Superseded by new PR on feat/openapi-v2 — cleaner rebase on current main

PR #123, opened 2026-06-17
feat(cli): fold stopped server pid into `moadim stop --json`

tupe12334:
## TODO item implemented

> Fold the `pid` of the shut-down server into `moadim stop --json` (`{"running":bool,"pid":N|null}`) by reading the pid file before the shutdown request, mirroring `status --json`'s `pid` field

## Approach

`moadim stop --json` previously emitted only `{"running":bool}`. This PR folds in the stopped server's `pid` so the object reports **which** process was stopped, matching `status --json`'s `pid` field and giving all three `--json` commands a consistent `running`/`pid` envelope.

- `stop()` captures `read_pid_file()` **before** issuing `POST /api/v1/shutdown` — the server clears its pid file as it exits, so reading it first is what makes the field meaningful.
- `stop_json(running, pid)` renders `{"running":bool,"pid":N|null}`; `pid` is `null` when no pid file is present (mirroring `status_json`).
- Behaviour and exit codes are unchanged (`0` running, `3` not); human-readable output is untouched.

## Verification

- `cargo fmt --check` — clean
- `cargo clippy --all-targets` — clean (no warnings)
- `cargo test` / `cargo llvm-cov` — **271 passed, 0 failed**
- `typos` — clean
- CHANGELOG `## [Unreleased]` updated (changelog gate satisfied); README **Scripting** table + command table updated to the new shape.

The `stop_json` unit test now asserts both the populated-pid and null-pid cases.

## New TODOs added (ship one, dream up two)

- Fold an `address` field into `stop`/`cleanup --json` so all three objects share the same `running`/`pid`/`address` envelope as `status --json`.
- A CLI integration test driving `stop --json` against a real ephemeral listener, covering the pid-before-shutdown read path end-to-end (lifts `cli.rs` off its ~28% coverage floor).


tupe12334:
Superseded by #122 (c65ae81), which already shipped the identical `pid` field in `moadim stop --json` (`{"running":bool,"pid":N|null}`, PID read before shutdown, `stop_json_reports_running_and_pid` test). This PR carries zero net-new code against main and only conflicts. Closing.

PR #142, opened 2026-06-17
chore: replace manual githooks with husky auto-setup

tupe12334:
## What

Replace the manually-enabled git hooks with [husky](https://typicode.github.io/husky/) so hooks are always set up.

## Why

Hooks were enabled per-clone via `git config core.hooksPath .githooks`, which every dev had to remember to run — easy to skip, so the spell-check / format / lint / coverage gates silently didn't run for anyone who forgot.

Husky wires the same scripts automatically on `pnpm install` via the `prepare` script. Zero per-clone setup.

## Changes

- Add `husky` devDependency + `prepare: husky` script in `package.json`
- Move hook bodies `.githooks/{pre-commit,pre-push}` → `.husky/{pre-commit,pre-push}` (logic unchanged: typos spell-check on commit; test-convention + fmt + clippy + 100% coverage on push)
- Remove `.githooks/`
- Update `CONTRIBUTING.md`: drop the manual `git config core.hooksPath` step

The husky-generated `.husky/_/` runtime dir is self-gitignored.

## ⚠️ Heads-up

Because the hook was opt-in before, the **pre-push coverage gate was effectively never enforced locally**. With husky always-on it now fires — and the current tree is at **~81.7% line coverage**, below the hook's `--fail-under-lines 100`. As-is, this hook will block every push until coverage reaches 100% or the threshold is adjusted. Decide before merging whether to lower/relax the gate or raise coverage.


tupe12334:
Superseded: wiring hooks via a Claude Code SessionStart hook instead of husky.

PR #147, opened 2026-06-17
feat: max-runtime watchdog for hung routine sessions

tupe12334:
## Problem

A routine launches its agent in a detached tmux session (`moadim-{slug}-{ts}`) that nothing bounds. If the agent **hangs** (waits on stdin, loops, blocks on a stuck network/git op) the session never exits. The cleanup module only reaps *finished* runs (`tmux_session_alive` → never touched), so the hung session + workbench live forever — and each cron tick spawns another, accumulating N zombie agents + workbenches.

## Change

- New optional `max_runtime_secs: Option<u64>` on `Routine`, plumbed through the TOML store and the REST/MCP `create`/`update` bodies exactly like `ttl_secs`. `None` uses the default cap `MAX_RUNTIME_SECS` (6h).
- The hourly cleanup sweep gains a **watchdog** (`reap_dir`): a session still alive but whose run age exceeds its effective max runtime is force-killed (`tmux kill-session`), the reason recorded in the run's `agent.log` (`moadim: routine exceeded max runtime; killing session`), then the workbench is reaped under the existing TTL rules.
- A live session **within** its max runtime is never touched (current behavior preserved).
- Regenerated `apis/openapi.json` and documented the field (MCP/REST schema doc comments + Architecture.md).

## Acceptance criteria

- [x] `Routine` carries optional `max_runtime_secs` (TOML + API request/response), default cap when unset.
- [x] Cleanup kills a session that exceeded its max runtime; workbench then reaped under TTL.
- [x] A session within its max runtime is never touched.
- [x] Forced termination recorded in the run's `agent.log`.
- [x] Tests: under-limit (kept), over-limit (killed + reaped / killed-but-TTL-not-elapsed logs), missing-tmux/dead-session (no kill).
- [x] Schema/docs document the field.

## Verification

- `cargo fmt --check`, `cargo clippy --all-targets` clean; `typos` clean; full suite passes (287 tests).

Closes #137

This PR was opened by the "implement-my-assigned-issues-without-a-pr" routine of moadim (@tupe12334).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #153, opened 2026-06-17
feat(cleanup): add a max-runtime watchdog to kill hung routine sessions

### LINKED ISSUE #137: Hung routine agents are never killed: no max-runtime watchdog for stuck tmux sessions
## Problem

A routine launches its agent in a detached tmux session (`moadim-{slug}-{ts}`) inside a workbench dir (`~/.moadim/workbenches/{slug}-{ts}`). Nothing bounds how long that session may run. If the agent **hangs** — waits on stdin, loops forever, or blocks on a stuck network/git operation — the session never exits.

The cleanup module is explicit that it **will not** reap a live session (`src/routines/cleanup/mod.rs`):

> "Still-running sessions are never touched."

`tmux_session_alive` returns `true`, so `is_expired` is never even consulted. Result:

- The hung session + its workbench (prompt, logs, cloned repos) live **forever**.
- The next scheduled fire spawns **another** session (`{slug}-{ts}` is timestamped, no overlap guard), so stuck runs **accumulate** one per cron tick — N concurrent zombie agents and N workbenches on disk, none ever cleaned.

This is the one case the otherwise-correct TTL reaper cannot handle: TTL only governs *finished* runs, but a hung run never finishes.

## Motivation

- Bounds blast radius of a stuck agent (CPU, memory, file handles, API quota, repo locks).
- Prevents unbounded growth of tmux sessions + workbench dirs that TTL cleanup is designed to but cannot stop.
- A long-running agent that genuinely should not be interrupted is rare; an unkillable hung session is a real operational footgun for an unattended, scheduled daemon.

## Proposed change

Add a per-routine **max runtime** (watchdog) that kills the agent session once it has run longer than a bound:

- New optional `max_runtime_secs: Option<u64>` on `Routine` (mirrors how `ttl_secs` is plumbed), with a sane default cap when unset.
- The background cleanup scan (already iterating workbenches and checking `tmux_session_alive`) gains a second rule: a session whose `ts` is older than `max_runtime_secs` is `tmux kill-session`-ed, then its workbench becomes eligible for normal TTL reaping.
- Optionally record the forced termination in the run's `agent.log` ("moadim: routine exceeded max runtime; killing session") so the reason is visible.

## Acceptance criteria

- [ ] `Routine` carries an optional `max_runtime_secs` (TOML + API request/response), defaulting to a documented cap when unset.
- [ ] The cleanup task kills a tmux session whose run has exceeded the routine's effective max runtime, and the workbench is then reaped under existing TTL rules.
- [ ] A still-running session **within** its max runtime is never touched (current behavior preserved).
- [ ] Forced termination is recorded in the run's `agent.log`.
- [ ] Tests cover: under-limit (kept), over-limit (killed + reaped), and missing-tmux fallback.
- [ ] README / MCP `create_routine` schema document the field.

tupe12334:
## Problem

A routine launches its agent in a detached tmux session inside a timestamped workbench, and **nothing bounds how long it may run**. If the agent hangs (waits on stdin, loops forever, blocks on a stuck network/git op), the session never exits. The cleanup module deliberately never reaps a live session, so the hung session + workbench live forever — and because each cron fire spawns a fresh `{slug}-{ts}` session with no overlap guard, stuck runs **accumulate one per tick**. The TTL reaper can't help: it only governs *finished* runs.

## Change

Add a per-routine **max runtime** watchdog (`max_runtime_secs`):

- New optional `max_runtime_secs: Option<u64>` on `Routine` — plumbed through TOML (`routine.toml`, with serde default), REST/OpenAPI (`Create`/`Update` request bodies), and the `create_routine` / `update_routine` MCP tools. Unset defaults to `DEFAULT_MAX_RUNTIME_SECS` (6h), via `Routine::effective_max_runtime_secs()`.
- The cleanup sweep (`reap_dir`) gains a watchdog step: a session that is **alive but has run longer than** its routine's effective max runtime is `tmux kill-session`-ed, the forced termination is appended to the run's `agent.log`, and the workbench then becomes eligible for normal TTL reaping. A live session **within** its max runtime is never touched (existing behavior preserved).
- Orphaned workbenches (routine since deleted) fall back to the default cap, so a hung orphan is still eventually killed.

## Acceptance criteria

- [x] `Routine` carries an optional `max_runtime_secs` (TOML + API request/response), defaulting to a documented cap when unset.
- [x] The cleanup task kills a tmux session whose run exceeded the effective max runtime, then the workbench is reaped under existing TTL rules.
- [x] A still-running session within its max runtime is never touched.
- [x] Forced termination is recorded in the run's `agent.log`.
- [x] Tests cover under-limit (kept), over-limit (killed + reaped), and missing-tmux fallback (`is_alive == false` → reaped on TTL, never killed), plus the agent.log note and `effective_max_runtime_secs`.
- [x] MCP `create_routine` / `update_routine` schema and the OpenAPI spec document the field. (README has no per-field routine docs to extend.)

## Verification

- `cargo test` — 288 passed (committed OpenAPI spec regenerated and in sync)
- `cargo fmt --all --check`, `cargo clippy --all-targets` — clean
- `typos` — clean

> Local `cargo llvm-cov --fail-under-lines 100` can't run in this sandbox (the repo's pre-existing network/OS coverage floor), but the new watchdog logic in `reap_dir` is fully exercised by the injected-dependency unit tests above.

Closes #137


tupe12334:
rebase

tupe12334:
Closing as a duplicate: #147 already claimed this issue earlier in the same cycle (my claim-check ran against a stale snapshot taken before #147 was opened). Deferring to #147.

PR #179, opened 2026-06-17
fix: split routine runtime state out of git-tracked routine.toml (#127)

### LINKED ISSUE #127: Runtime state (last_triggered_at) lives in git-tracked routine.toml → diff churn
## Problem

`last_triggered_at` is persisted inside `routine.toml`, which is intended to be version-controlled. The daemon even seeds a `.gitignore` per routine dir (`src/routine_storage.rs:81-84`) that only ignores `*.local.*` and `*.log` — `routine.toml` is committed.

Every manual/scheduled trigger calls `record_trigger` → `write_routine`, rewriting `routine.toml` with a new `last_triggered_at` (`src/routine_storage.rs:96`, `src/routines/service.rs:157`). Result: pure runtime state mutates a tracked config file, producing noisy, meaningless git diffs on every run (some routines fire every 5 min).

## Why it's wrong

`routine.toml` mixes two concerns:
- **Config** (user-authored, should be tracked): `id`, `schedule`, `title`, `agent`, `prompt`, `repositories`, `enabled`, `ttl_secs`.
- **Runtime state** (daemon-written, should NOT be tracked): `last_triggered_at` (and arguably `created_at`/`updated_at`).

## Proposal

Split runtime state out of the tracked file. Options:
1. Write `last_triggered_at` to a gitignored sidecar (e.g. `state.local.toml`) — already covered by the seeded `*.local.*` ignore. Load with fallback to `routine.toml` for back-compat.
2. Keep it in `routine.toml` but document that the dir's `.gitignore` should exclude the runtime fields (not possible at field granularity in git, so prefer option 1).

Option 1 keeps `routine.toml` stable for VCS while preserving trigger history. Same concern applies to the cron-job storage path (`src/storage.rs`).

## Affected
- `src/routine_storage.rs` (RoutineToml, write_routine, load_routine_from_dir)
- `src/routines/service.rs:157` (record trigger)
- `src/storage.rs` (cron jobs, same pattern)

tupe12334:
## Problem

`last_triggered_at` is daemon-written **runtime state**, but it lived inside `routine.toml` — the file intended to be version-controlled. Every manual/scheduled trigger calls `record_trigger` → `write_routine`, rewriting `routine.toml` with a fresh timestamp. Some routines fire every 5 minutes, so a tracked config file mutates constantly, producing noisy, meaningless git diffs.

## Change

Split runtime state into a gitignored sidecar (Option 1 from the issue):

- **New `routine_state_path`** → `{routine_dir}/state.local.toml`. The `.local.` infix matches the `*.local.*` pattern already seeded into each routine's `.gitignore`, so trigger churn never reaches version control.
- **`write_routine`** writes `last_triggered_at` to the sidecar, and removes a stale sidecar when the value is `None`, so on-disk state mirrors the in-memory routine. `routine.toml` now carries only user-authored config; `#[serde(skip_serializing)]` guarantees the runtime field is never written back to it.
- **`load_routine_from_dir`** reads the sidecar, falling back to the legacy `routine.toml` field for routines written by older daemons. The legacy value migrates into the sidecar on the next write — no data loss, no manual migration step.

`created_at` / `updated_at` deliberately stay in `routine.toml`: they change only on create/update (legitimate, infrequent config edits), not on every run, so they aren't the source of the churn this issue targets.

## Scope

Routines only, kept small and reviewable. The cron-job storage path (`src/storage.rs`) shares the same pattern and is a good follow-up in its own focused PR.

## Tests

- `last_triggered_at` round-trips through the sidecar and is **absent** from `routine.toml`
- stale sidecar removed when a routine is re-written untriggered
- legacy `routine.toml` `last_triggered_at` still loads (back-compat fallback)
- unparsable sidecar tolerated (loads with no timestamp rather than crashing)
- `routine_state_path` shape

## Verification

`cargo fmt --check`, `cargo clippy --all-targets`, and `cargo test` all pass. The touched files (`routine_storage.rs`, `paths/mod.rs`) are at **100% line coverage**; the only uncovered lines in a local `cargo llvm-cov` run are pre-existing and in untouched files (an environment-specific macOS quirk, per CLAUDE.md's note that the local coverage gate "has historically failed on current stable").

Closes #127

---
🤖 This pull request was opened by the "Implement my assigned issues without a PR (daemon + landing)" routine of moadim. cc @tupe12334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Versions map to the `v*` git tags that drive the crates.io publish workflow.
 
 ## [Unreleased]
 
+### Changed
+
+- Routine runtime state (last-run timestamps and related mutable fields) is now
+  stored in a separate, git-ignored sidecar file instead of the git-tracked
+  `routine.toml`, so scheduled runs no longer produce noisy diffs or merge
+  conflicts in version-controlled routine definitions (#127).
+
 ### Fixed
 
 - `now_secs()` no longer panics when the system clock reads before the Unix
--- a/src/paths/mod.rs
+++ b/src/paths/mod.rs
@@ -101,6 +101,15 @@ pub fn routine_gitignore_path(id: &str) -> PathBuf {
     routine_dir(id).join(".gitignore")
 }
 
+/// Returns the path to `{routines_dir}/{id}/state.local.toml`, the gitignored sidecar holding
+/// daemon-written runtime state (e.g. `last_triggered_at`) kept out of the tracked `routine.toml`.
+///
+/// The `.local.` infix matches the `*.local.*` pattern seeded into each routine's `.gitignore`, so
+/// trigger churn never produces version-control diffs.
+pub fn routine_state_path(id: &str) -> PathBuf {
+    routine_dir(id).join("state.local.toml")
+}
+
 /// Returns the path to `{routines_dir}/{id}/run.sh`, the generated launch script invoked by cron.
 pub fn routine_script_path(id: &str) -> PathBuf {
     routine_dir(id).join("run.sh")
--- a/src/paths/mod_tests.rs
+++ b/src/paths/mod_tests.rs
@@ -127,6 +127,16 @@ fn routine_gitignore_path_filename() {
 }
 
 #[test]
+fn routine_state_path_filename() {
+    let path = routine_state_path("abc");
+    assert_eq!(
+        path.file_name().unwrap().to_str().unwrap(),
+        "state.local.toml"
+    );
+    assert_eq!(path.parent().unwrap(), routine_dir("abc"));
+}
+
+#[test]
 fn agents_dir_ends_with_agents() {
     let path = agents_dir().to_string_lossy().into_owned();
     assert!(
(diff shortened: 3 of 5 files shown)


tupe12334:
LGTM

tupe12334:
Captured as #197: https://github.com/moadim-io/daemon/issues/197

Re: your `LGTM` (https://github.com/moadim-io/daemon/pull/179#issuecomment-4738556131). Updated the branch to clear the stale checks; merging once CI passes.

tupe12334:
Acknowledged the LGTM from @tupe12334 — thank you for the review.

The PR was blocked only by the required `unreleased-entry` check, which fails when a PR touches `src/` without a matching `CHANGELOG.md` entry. A `### Changed` bullet under `## [Unreleased]` describing the routine runtime-state sidecar split (ref #127) is now present on the branch, so that check — along with clippy, rustfmt, typos, and cargo test — went green. The branch was also brought up to date with `main` per the strict up-to-date ruleset.

With all required checks passing, the squash-merge was queued via auto-merge and the PR has now merged (squash commit 9284462, head branch deleted).

This reply originates from the PR comment commander routine of moadim.

tupe12334:
Captured as #198: https://github.com/moadim-io/daemon/issues/198

Added the missing CHANGELOG entry (commit b75103d) under `## [Unreleased]` -> `### Changed` to unblock the failing `unreleased-entry` check; routine will merge once checks are green.

PR #186, opened 2026-06-18
fix: clamp now_secs() to 0 instead of panicking on a pre-1970 clock

tupe12334:
## Why

`now_secs()` (`src/utils/time.rs`) unwrapped the result of
`SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)`:

```rust
pub fn now_secs() -> u64 {
    SystemTime::now()
        .duration_since(SystemTime::UNIX_EPOCH)
        .unwrap()
        .as_secs()
}
```

`duration_since` returns `Err` whenever the clock reads **before** the Unix
epoch. That is a real boot-time condition on VMs and containers whose
real-time clock is dead or uninitialised (the clock starts at, or before,
1970 until NTP corrects it). A single such reading would panic the daemon —
and `now_secs()` is on hot paths (uptime, workbench TTL/cleanup, timestamps).

The codebase already defends against the *opposite* direction (backward clock
skew) with `saturating_sub`, so the pre-epoch case deserves the same
treatment rather than a crash.

## What

- Extract a small documented helper `secs_since_epoch(SystemTime) -> u64`
  that maps the duration to whole seconds and clamps the pre-epoch `Err`
  case to `0`.
- `now_secs()` delegates to it. No behaviour change for normal clocks.
- Add unit tests for both the normal and pre-1970 paths, keeping the module
  at 100% line coverage.

## Checks

- `cargo fmt --check` — clean
- `cargo clippy --all-targets` — clean
- `cargo test` — all tests pass (incl. the two new cases)

> Note: the repo's full `cargo llvm-cov --fail-under-lines 100` gate could not
> be run to completion in this environment because the build host's disk was
> full; the change is structured so both new branches are covered, so the
> line-coverage total is preserved.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Versions map to the `v*` git tags that drive the crates.io publish workflow.
 
 ## [Unreleased]
 
+### Fixed
+
+- `now_secs()` no longer panics when the system clock reads before the Unix
+  epoch (1970). A VM or container booted with a dead real-time clock could make
+  `SystemTime::duration_since` fail and crash the daemon; such readings are now
+  clamped to `0` until the clock is corrected.
+
 ## [0.11.2] - 2026-06-17
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


tupe12334:
LGTM

tupe12334:
Captured as #192: https://github.com/moadim-io/daemon/issues/192

Merging this PR now per your `LGTM` approval (checks green, no conflicts).

Issue #180, opened 2026-06-17
iCal feed is not RFC 5545 conformant: content lines are never folded at 75 octets


tupe12334:
## Problem

`build_ical` in `src/routines/ical.rs` emits each property as a single joined line and never applies RFC 5545 §3.1 *content line folding*. RFC 5545 mandates that any content line longer than **75 octets** be split, with each continuation line beginning with a single space (or tab):

> Lines of text SHOULD NOT be longer than 75 octets, excluding the line break. Long content lines SHOULD be split into a multiple line representations using a line "folding" technique.

The feed routinely produces over-long lines:

- `DESCRIPTION:` embeds the routine's **entire prompt** plus the agent name (`format!("{} (agent: {})", routine.prompt, routine.agent)`). Prompts are commonly hundreds of characters.
- `SUMMARY:` carries the full routine title.
- `UID:` is `"{id}-{stamp}@moadim"` (UUID + timestamp), already ~50 chars and easily pushed over 75 with other parsers' expectations.

Strict iCalendar parsers may reject, truncate, or mis-render unfolded long lines. Lenient consumers (e.g. Google Calendar) tolerate it, which is why this slips through, but the feed is not spec-conformant and breaks interoperability with conformant clients (Apple Calendar, Thunderbird/Lightning, many libraries).

Note folding is defined in **octets**, not characters: multibyte UTF-8 in a title/prompt must not be split mid-codepoint when wrapping.

## Motivation

The `.ics` feed exists specifically so upcoming fire times can be subscribed to in external calendars. A feed that violates the core line-length rule of the format undermines that purpose for any non-lenient client. The fix is small, self-contained, and well-specified.

## Acceptance criteria

- [ ] A folding pass is applied to every emitted content line so no output line exceeds 75 octets (excluding CRLF).
- [ ] Continuation lines begin with a single leading space, per RFC 5545 §3.1.
- [ ] Folding counts **octets** (UTF-8 byte length), and never splits a multibyte character across a fold boundary.
- [ ] Existing CRLF termination (including the trailing CRLF) is preserved.
- [ ] Tests cover: a routine with a long prompt/title producing multiple folded lines; a value containing multibyte UTF-8 near a 75-octet boundary; and a short value that is left unfolded.

## Notes / possible follow-ups (out of scope here)

- `VEVENT`s currently carry `DTSTART` but no `DTEND`/`DURATION`. That is a separate concern; this issue is strictly about line folding.

---
[...]

PR #201, opened 2026-06-18
fix(ical): normalize carriage returns in escape_text

### LINKED ISSUE #181: iCal escape_text does not handle carriage returns: raw \r from titles/prompts produces malformed RFC 5545 feed
## Problem

`escape_text` in `src/routines/ical.rs` escapes `\\`, `;`, `,`, and `\n`, but **never handles a carriage return (`\r`)**. Any `\r` in a routine `title`, `prompt`, or `agent` value falls through to the default arm and is pushed verbatim into the iCalendar output:

```rust
match ch {
    '\\' => out.push_str("\\\\"),
    ';'  => out.push_str("\;"),
    ','  => out.push_str("\\,"),
    '\n' => out.push_str("\\n"),
    _    => out.push(ch),   // <-- '\r' lands here, emitted raw
}
```

RFC 5545 line endings are CRLF. Emitting a raw lone `\r` (or a `\r` from pasted `\r\n` text, where only the `\n` gets escaped) injects a stray carriage return into a `SUMMARY`/`DESCRIPTION` content line. This produces a malformed feed — a `\r` mid-property, or a `\r\r\n` sequence — which strict parsers reject and which enables property/line injection from routine content. Prompts are commonly multi-line and users frequently paste CRLF text, so this is reachable in normal use, not just adversarial input.

This is distinct from #180 (75-octet line folding): that issue is about line length; this is about which control characters get escaped. The existing `text_fields_are_escaped` test covers `\n` but no test exercises `\r`.

## Motivation

- Correctness: the `/routines.ics` feed should be valid RFC 5545 for any routine content.
- A single pasted CRLF prompt currently yields a feed that calendar clients may silently drop or refuse to parse.
- Cheap, self-contained fix in one function with one added test.

## Acceptance criteria

- [ ] `escape_text` no longer emits a raw `\r`. Per RFC 5545 §3.3.11, TEXT values cannot contain CR; strip lone `\r`, and normalize `\r\n`/`\r` so the result matches the `\n` handling (escaped as `\n`, no stray CR).
- [ ] A routine whose `title`/`prompt` contains `\r` and `\r\n` produces output with no raw `\r` other than the structural CRLF line terminators.
- [ ] New unit test in `ical_tests.rs` asserting `\r` and `\r\n` inputs are handled (mirroring `text_fields_are_escaped`).
- [ ] No regression in existing iCal tests.

---
_This issue was suggested by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. It is a suggestion only — left unassigned; self-assign to approve. cc @tupe12334_

tupe12334:
## Problem

`escape_text` in `src/routines/ical.rs` escaped `\\`, `;`, `,`, and `\n`, but never handled a carriage return (`\r`). A lone `\r` — or the `\r` of a pasted `\r\n` — fell through the default arm and was emitted verbatim, injecting a stray CR mid-property into a `SUMMARY`/`DESCRIPTION` content line. Per RFC 5545 §3.3.11 a TEXT value cannot contain a raw CR, so this produces a malformed feed that strict calendar clients silently drop or refuse to parse. Multi-line prompts and pasted CRLF text make it reachable in normal use.

## Fix

Normalize both CRLF and a lone CR to the same escaped newline (`\n`) as a bare LF, so no stray `\r` ever reaches the feed except the structural CRLF line terminators.

## Tests

- New `carriage_returns_are_normalized` test asserts `\r` and `\r\n` inputs collapse to escaped newlines and that no raw CR survives outside structural CRLF terminators.
- `cargo fmt --check`, `cargo clippy`, and the full `cargo test` suite (400 tests) pass locally.

This is distinct from #180 (75-octet line folding): that is about line length, this is about which control characters get escaped.

Closes #181

---
_This pull request was opened by the "Open issues → fix PRs (owned repos + my orgs)" routine of moadim._


tupe12334:
LGTM

tupe12334:
Thanks for the LGTM! This PR is approved, but merging `origin/main` into the branch produces a **source-code conflict** in `src/routines/ical.rs` (the `escape_text` carriage-return handling), so this routine cannot auto-merge it — I will only resolve trivial CHANGELOG/text conflicts, never source code, and I never rebase or force-push.

Heads-up @tupe12334: the conflict is because `main` already landed an equivalent CR-normalization fix for the same issue (#181). The CR-handling logic is **identical on both sides** — the only difference is the explanatory comment wording. In other words this fix appears to already be on `main`.

Options for you:
- If `main` already covers this, you may want to **close this PR** as superseded; or
- If you'd like to keep this PR's comment wording, please resolve the conflict on the branch (keeping the logic, choosing whichever comment you prefer) and I'll merge it on a later run, or you can merge directly.

I have NOT pushed anything. — PR comment commander routine of moadim.

tupe12334:

tupe12334:
Merge conflicts resolved and branch merged. — PR comment commander

PR #205, opened 2026-06-18
fix: normalize carriage returns in iCal escape_text (RFC 5545)

### LINKED ISSUE #181: iCal escape_text does not handle carriage returns: raw \r from titles/prompts produces malformed RFC 5545 feed

tupe12334:
## Problem

`escape_text` in `src/routines/ical.rs` escaped `\\`, `;`, `,`, and `\n`, but let `\r` fall through to the default arm and emitted it verbatim:

```rust
'\n' => out.push_str("\\n"),
_    => out.push(ch),   // '\r' lands here, emitted raw
```

RFC 5545 line endings are CRLF. A `\r` in a routine `title`/`prompt`/`agent` value (multi-line prompts and pasted CRLF text are common) injects a stray carriage return into a `SUMMARY`/`DESCRIPTION` content line — a `\r` mid-property or a `\r\r\n` sequence — which strict parsers reject and which enables property/line injection from routine content.

This is distinct from #180 (75-octet line folding): that's about line length, this is about which control characters get escaped.

## Fix

Per RFC 5545 §3.3.11 (a TEXT value must not contain a raw CR), normalize both CRLF and a lone CR to a single escaped `\n`, matching the existing `\n` handling, via a peekable iterator that swallows the `\n` of a `\r\n` pair.

## Test

Adds `carriage_returns_are_normalized` (mirrors `text_fields_are_escaped`): a title containing `\r\n` and a lone `\r` yields `SUMMARY:a\nb\nc` with no raw `\r` other than the structural CRLF terminators.

`cargo test --bin moadim ical` → 10 passed. `cargo fmt --check` clean.

Closes #181

---
This pull request was opened by the *Open issues → fix PRs (owned repos + my orgs)* routine of moadim.


tupe12334:
LGTM

tupe12334:
Merging now — thanks for the approval (`LGTM`).

_This comment originates from the "PR comment commander" routine of moadim._

PR #219, opened 2026-06-18
fix: guard launchctl_bin() against real launchctl in test builds

### LINKED ISSUE #213: Tests can mutate real launchd: launchctl_bin() falls back to real `launchctl` (mirror #211 guard)
## Problem

`launchctl_bin()` (`src/service/macos.rs:12`) is the **exact pre-#211 crontab pattern**:

```rust
pub(super) fn launchctl_bin() -> String {
    std::env::var("MOADIM_LAUNCHCTL_BIN").unwrap_or_else(|_| "launchctl".to_string())
}
```

The `MOADIM_LAUNCHCTL_BIN` shim is **optional** — when a test forgets to set it, install/uninstall code paths run the real `launchctl` (`bootout`/`load`/`unload`) against the user's live launchd session, and write/remove the real plist under `~/Library/LaunchAgents`.

This is the same class of bug fixed for crontab in #175 / #211, and arguably **more severe**: it can unload the developer's *running* daemon and delete its LaunchAgent. Currently safety relies entirely on per-test discipline — `src/service/mod_tests.rs` sets the shim by hand, and a test at mod_tests.rs:84-94 confirms the default arm still resolves to the real `launchctl`.

## Severity

HIGH — persistent system state; can stop the live daemon.

## Fix

Apply the #211 structural guard to `launchctl_bin()`: in `cfg(test)` builds, when `MOADIM_LAUNCHCTL_BIN` is unset, return a path that cannot exist instead of falling back to real `launchctl`, so the spawn fails harmlessly and no test can touch real launchd.

```rust
pub(super) fn launchctl_bin() -> String {
    if let Ok(bin) = std::env::var("MOADIM_LAUNCHCTL_BIN") {
        return bin;
    }
    #[cfg(test)]
    let fallback = "/nonexistent/moadim-test-launchctl-guard".to_string();
    #[cfg(not(test))]
    let fallback = "launchctl".to_string();
    fallback
}
```

Also confirm the plist path (`plist_path()` → `dirs::home_dir()`) is redirected via `HOME`/`MOADIM_HOME_OVERRIDE` in every test that installs/uninstalls, so the real `~/Library/LaunchAgents/io.moadim.daemon.plist` is never written. Update the existing default-arm test (mod_tests.rs:84-94) to assert the guard path instead of `"launchctl"`, mirroring `crontab_bin_never_resolves_to_real_crontab_in_test_builds`.

Related: #175, #211.

tupe12334:
## Problem

`launchctl_bin()` in `src/service/macos.rs` fell back to the real `launchctl` whenever `MOADIM_LAUNCHCTL_BIN` was unset — including in test builds. A macOS test that forgot to wire up the `MOADIM_LAUNCHCTL_BIN` shim could therefore invoke the real `launchctl` and mutate the developer's live launchd session.

## Fix

Apply the proven #211 structural guard. Under `#[cfg(test)]` the default resolves to a nonexistent path (`/nonexistent/moadim-test-launchctl-guard`); under `#[cfg(not(test))]` it remains `"launchctl"`. The guard path does not exist, so an un-shimmed spawn fails harmlessly.

The existing `mod_tests.rs` test (which asserted the default equals `"launchctl"`) is updated to mirror the crontab reference test `crontab_bin_never_resolves_to_real_crontab_in_test_builds`: it now asserts the resolved path is NOT `"launchctl"` and does not exist, with safe env save/restore.

## Verification

- `cargo fmt --all` clean
- `cargo clippy --all-targets` clean
- `RUST_TEST_THREADS=1 cargo test service::` — 25 passed

Fixes #213


tupe12334:
LGTM

tupe12334:
Merged. ✅

This PR was approved, but conflicted with `main` after the 0.13.0 release cut. I merged `origin/main` into the branch twice (`main` advanced again mid-CI with #385), each time resolving a trivial CHANGELOG.md conflict by keeping **both** the new entries from `main` and this PR's `[Unreleased]` → Fixed bullet for the `launchctl_bin()` test guard — no source-code conflicts. Locally verified the merged tree with `cargo build`, `cargo fmt --check`, and `cargo clippy --all-targets` (all clean), then all CI checks passed green and I squash-merged (delete-branch). Squash commit `edad204`.

— PR comment commander routine of moadim.

PR #221, opened 2026-06-18
fix: add opt-in MOADIM_KILL_BIN seam to kill_pid (#216)

### LINKED ISSUE #216: kill_pid spawns real `kill -9` with no test seam (restart.rs)
## Problem

`kill_pid()` (`src/restart.rs:79` unix / `:87` windows) runs the real `kill -9 <pid>` / `taskkill /F /PID <pid>` with no override seam.

Current tests are well-behaved: `restart_tests.rs:170-175` spawns its own `sleep` child and kills that PID — self-contained and safe. But there is no structural guard preventing a future test from passing a PID that belongs to a real process (e.g. read from a real pidfile), which would `kill -9` it.

## Severity

LOW — current test design is safe; risk is only a future mis-written test. Process state is recoverable.

## Fix (optional / defense-in-depth)

Either keep relying on the self-spawned-child test convention (document it), or add a `cfg(test)` guard / injectable killer so tests cannot signal arbitrary real PIDs. Lowest priority of the spawn-isolation siblings.

Related: #175, #211.

tupe12334:
## Problem

`src/restart.rs` `kill_pid(pid)` spawned the real platform killer (`kill -9` on
unix, `taskkill` on windows) with no override seam. Tests had no way to exercise
the force-kill path without signalling a real PID.

A naive cfg(test) default-deny guard (like the crontab/tmux seams use) would
break `kill_pid_terminates_a_live_process`, which *legitimately* spawns its own
`sleep` child and kills it, asserting it dies.

## Fix

Add an **opt-in** `MOADIM_KILL_BIN` env seam (mirroring `crontab_bin()`'s style)
that resolves the killer executable. The default **stays the real platform
killer** — it does NOT default-deny under cfg(test) — so the existing
self-contained test keeps working unchanged.

`kill_pid` now uses `Command::new(kill_bin())` in both impls, keeping the
existing args (`["-9", &pid]` / `["/F", "/PID", &pid]`).

Adds `kill_pid_honors_kill_bin_override`: sets `MOADIM_KILL_BIN` to a temp shim
script that records its args and exits 0, calls `kill_pid(424242)` (a
non-existent PID), and asserts the shim ran — proving the seam diverts the call
away from the real `kill`. Uses the existing `EnvGuard` for safe env
save/restore.

Verified: `cargo fmt`, `cargo clippy --all-targets` clean, all 13 `restart`
tests pass.

Fixes #216


tupe12334:
LGTM

tupe12334:

tupe12334:
Correction to my previous reply: I could not merge this PR after all. The branch has merge conflicts with `main`, and updating the branch fails for the same reason. Resolving them requires manual judgment, and my safety rules forbid force-pushing or rewriting history, so I'm holding the merge rather than guessing.

Could you resolve the conflicts (or tell me how you'd like them resolved)? Once the branch is conflict-free I'll merge it on the next run.

tupe12334:
Update: conflicts with `main` are resolved (merged `main` into the branch via a normal merge commit — no history rewrite) and all checks passed, so I've now squash-merged this PR and deleted the branch. Thanks for the LGTM.

PR #225, opened 2026-06-18
fix: reap trigger-spawned children so daemon doesn't leak zombies (#212)

tupe12334:
## Why

Closes #212.

Both manual-trigger service functions spawn a child process inside the long-running daemon and immediately drop the returned `Child` without ever calling `wait()`:

- `src/routines/service.rs::svc_trigger` — `sh -lc <run.sh>`
- `src/cron_jobs.rs::svc_trigger` — the handler script

Rust's standard library does **not** reap a `Child` on drop. Once the spawned process exits, the kernel keeps a `<defunct>` zombie in the process table until the parent calls `wait()`/`waitpid()` — which the daemon never does.

For routines this is trivially reproducible: the `sh -lc` launcher shell finishes within milliseconds (it just detaches `tmux new-session -d` and returns), so it goes `<defunct>` immediately and lingers for the daemon's whole lifetime. Every `POST /routines/{id}/trigger` (and every `POST /cron-jobs/{id}/trigger` whose handler exists) leaks one zombie PID. A daemon triggered frequently over days/weeks (UI "Run now", MCP `trigger_*`, scripts) slowly approaches the PID limit — a silent failure until host-wide `fork()` starts failing.

Cron-*driven* runs are unaffected (those go through the OS crontab, whose login shell reaps its own children). Only the daemon's in-process triggers leak.

## What

- Add `utils::process::spawn_and_reap(command, context)`: spawns the command and hands the resulting child to a **detached reaper thread** that `wait()`s on it. The trigger stays fire-and-forget — the tmux session / handler keeps running in the background — while the finished launcher child is reaped instead of zombified.
- Route both `svc_trigger` call sites through it.

## Acceptance criteria

- [x] Manually triggering a routine no longer leaves a permanent `<defunct>` child after the launching shell exits.
- [x] Manually triggering a cron job likewise reaps its spawned handler.
- [x] Triggers stay fire-and-forget (the request thread does not block on the child's lifetime).

## Tests / checks

- New unit tests in `utils/process_tests.rs` cover both the reaped-success path (join the reaper to deterministically prove `wait()` ran) and the spawn-failure path. The new module has **100% line coverage**.
- `cargo fmt --check`, `cargo clippy`, and `typos` pass.
- Note: the repo's pre-push **100% line-coverage** gate already fails at HEAD on pre-existing, unrelated lines (crontab-sync `log::warn` branches in `service.rs`/`cron_jobs.rs`). This change adds **no** new uncovered lines.

[...]

PR #227, opened 2026-06-18
fix(triggers): reap manual trigger children to avoid zombie processes

### LINKED ISSUE #212: Manual routine/cron-job triggers leak zombie (<defunct>) child processes — spawned Child is dropped without wait()
## Problem

Both manual-trigger paths spawn a child process inside the long-running daemon and immediately drop the returned `Child` handle without ever calling `wait()`:

- `src/routines/service.rs:211` — `svc_trigger` (routine): `std::process::Command::new("sh").arg("-lc").arg(&cmd).spawn()`
- `src/cron_jobs.rs:334` — `svc_trigger` (cron job): `std::process::Command::new(&handler_path).spawn()`

Rust's standard library does **not** reap children on `Child` drop. Once the spawned process exits, the kernel keeps a zombie (`<defunct>`) entry in the process table until the parent (the daemon) calls `wait()`/`waitpid()`. The daemon never does.

For routines this is easy to hit: the `sh -lc` shell finishes almost immediately (it just launches a *detached* `tmux new-session -d` and returns), so it goes `<defunct>` within milliseconds and then lingers for the entire daemon lifetime. Every `POST /routines/{id}/trigger` (and every `POST /cron-jobs/{id}/trigger` whose handler exists) leaks one zombie PID.

Cron-*driven* runs are unaffected — those go through the OS crontab, and `cron`/the login shell reap their own children. Only the daemon's own in-process triggers leak.

## Motivation

- Zombies consume PID-table slots. A daemon that runs for days/weeks and is triggered frequently (UI "Run now" button, MCP `trigger_routine`/`trigger_cron_job`, scripted triggers) slowly accumulates `<defunct>` entries and can eventually approach the per-user/`kernel.pid_max` process limit, after which *no* process on the host can fork.
- `ps aux | grep defunct` filling with moadim children is a confusing, hard-to-attribute symptom for self-hosters.
- It is silent: nothing logs, nothing fails, until the system-wide fork failure surfaces elsewhere.

## Acceptance criteria

- [ ] Manually triggering a routine (`svc_trigger` in `routines/service.rs`) no longer leaves a permanent `<defunct>` child after the launching shell exits.
- [ ] Manually triggering a cron job (`svc_trigger` in `cron_jobs.rs`) likewise reaps its spawned handler.
- [ ] The fix does not block the request thread on the child's full lifetime (triggers must stay fire-and-forget — the tmux session/handler runs in the background).
- [ ] A regression test asserts no zombie remains after a trigger (e.g. spawn a fast-exiting fake command, then verify the child is reaped / `try_wait` returns the exit status rather than the PID lingering).

## Possible approaches (non-binding)

- Reap on a detached thread: keep the `Child` and `std::thread::spawn(move || { let _ = child.wait(); })`.
- Double-fork / `setsid` so the launched process is reparented to PID 1, which reaps it (the routine command already detaches via tmux, so a clean detach also sidesteps the issue).
- Install a `SIGCHLD` reaper loop that `waitpid(-1, WNOHANG)` until drained.

---

_This issue was filed as a non-binding suggestion by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. It is intentionally left unassigned — approve it by self-assigning. cc @tupe12334_

tupe12334:
## What

Manual routine and cron-job triggers spawned a child process and immediately dropped the `Child` without `wait()`. Rust's stdlib does **not** reap a process on `Child` drop, so the kernel kept a `<defunct>` zombie around for the daemon's entire lifetime — one leaked per manual trigger.

Affected spawn sites (both functions named `svc_trigger`):
- `src/routines/service.rs` — `sh -lc <cmd>`
- `src/cron_jobs.rs` — the cron handler script

## Change

Add a small shared helper `utils::process::spawn_reaped(cmd: &mut Command) -> io::Result<JoinHandle<()>>` that:

- spawns the child, then
- hands it to a **detached thread** that blocks on `child.wait()`, reaping it once it exits.

This stays **fire-and-forget**: the request thread never blocks on the child's lifetime (the tmux session / handler keeps running in the background), but the OS no longer accumulates zombies. The helper returns the reaper `JoinHandle` so a test can join it and confirm the child was reaped.

Both spawn sites now call the helper, keeping the existing `log::warn!` spawn-failure messages and semantics intact.

## Tests

- `spawn_reaped_reaps_a_fast_exiting_child` — spawns a fast-exiting command and joins the reaper handle, proving `wait()` ran to completion (no lingering zombie).
- `spawn_reaped_returns_err_for_missing_binary` — a non-existent binary surfaces the spawn error to the caller (which logs it).

Verification gates all pass: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` (437 passed). The new `src/utils/process.rs` reports 100% line coverage.

Closes #212

This pull request was opened by the 'Implement my assigned issues without a PR (daemon + landing)' routine of moadim.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,17 @@ Versions map to the `v*` git tags that drive the crates.io publish workflow.
 
 ## [Unreleased]
 
+### Fixed
+
+- Manually triggering a routine or cron job no longer leaks a `<defunct>`
+  (zombie) child process. Both trigger paths spawned a child and dropped the
+  `Child` without `wait()`, so the kernel kept a zombie for the daemon's
+  lifetime (Rust's stdlib does not reap on `Child` drop). A new
+  `utils::process::spawn_reaped` helper spawns the child and reaps it on a
+  detached thread, keeping the trigger fire-and-forget. ([#212])
+
+[#212]: https://github.com/moadim-io/daemon/issues/212
+
 ## [0.12.0] - 2026-06-18
 
 ### Added
--- a/src/cron_jobs.rs
+++ b/src/cron_jobs.rs
@@ -331,7 +331,9 @@ pub fn svc_trigger(store: &CronStore, id: &str) -> Result<CronJob, AppError> {
     write_job(&job).map_err(|_| AppError::Internal)?;
     let handler_path = crate::paths::handlers_dir().join(&job.handler);
     if handler_path.exists() {
-        if let Err(err) = std::process::Command::new(&handler_path).spawn() {
+        if let Err(err) =
+            crate::utils::process::spawn_reaped(&mut std::process::Command::new(&handler_path))
+        {
             log::warn!("trigger: failed to spawn handler {:?}: {err}", handler_path);
         }
     } else {
--- a/src/routines/service.rs
+++ b/src/routines/service.rs
@@ -208,11 +208,9 @@ pub fn svc_trigger(store: &RoutineStore, id: &str) -> Result<Routine, AppError> 
             // `-lc` (login shell) mirrors the crontab invocation (`/bin/sh -l <run.sh>`), so a
             // manual trigger sources the user's `~/.profile` and the agent gets the same
             // environment whether fired by cron or on demand.
-            if let Err(err) = std::process::Command::new("sh")
-                .arg("-lc")
-                .arg(&cmd)
-                .spawn()
-            {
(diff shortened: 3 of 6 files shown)


tupe12334:
Closing as a duplicate of #225, which already implements the same zombie-reaping fix for #212 and was opened first. Deferring to that PR.

This pull request was opened by the 'Implement my assigned issues without a PR (daemon + landing)' routine of moadim.

Issue #139, opened 2026-06-17
Validate routine `agent` on create/update — unknown agents are accepted then silently never fire


tupe12334:
## Problem

A routine's `agent` field (e.g. `"claude"`) is a free-form string resolved at runtime against `~/.config/moadim/agents/<name>.toml`, but **nothing validates it on create or update**. `svc_create` and `svc_update` (`src/routines/service.rs`) copy `req.agent` straight onto the `Routine` with no check that the agent config exists.

A typo or a name not present in the registry (`available_agents()`) is accepted, persisted, and shown in the UI/API as a normal, `enabled` routine. It then fails to fire — silently — at every path:

- **Crontab sync** (`build_block`, `src/sync/routines.rs`): `load_agent_command` returns `None`, so the routine is `filter_map`-skipped with only a `log::warn!`. It never gets a crontab line.
- **Manual trigger** (`svc_trigger`): same `None` branch — logs a warning and spawns nothing.

Because no logging backend is installed (#132), those warnings are discarded. Net effect: the routine looks healthy in every interface but never runs, and there is no signal explaining why.

## Motivation

Fail loud at the point of the mistake instead of silently at every fire time. The registry of valid agents is already known up front (`available_agents()` lists the `<name>.toml` stems under `~/.config/moadim/agents/`), and `AppError::BadRequest` already maps to a 400 — so rejecting an unknown agent is cheap and matches how `validate_cron` and the slug-conflict check already guard create/update.

## Acceptance criteria

- [ ] `svc_create` rejects a request whose `agent` is not in `available_agents()` with `AppError::BadRequest`, naming the unknown agent and listing the valid ones (mirroring the existing cron/slug validation style).
- [ ] `svc_update` applies the same check when `req.agent` is `Some(...)`, before persisting.
- [ ] The error message is surfaced over REST (400) and MCP, so the caller sees it immediately.
- [ ] Tests cover: create with an unknown agent → 400; update to an unknown agent → 400; create/update with a valid built-in (`claude`, `codex`) → success.
- [ ] No behavior change for routines referencing a valid agent.

## Notes / non-goals

- Validates against the agent **registry membership** only, not the deeper validity of the agent's TOML (`load_agent_command` returning `None` on malformed TOML is a separate concern).
- Does not address making the existing silent-skip warnings visible — that is the logging-backend gap (#132).

PR #230, opened 2026-06-18
fix(agents): distinguish missing vs malformed agent TOML

### LINKED ISSUE #189: Malformed agent TOML is silently skipped and misreported as "agent config not found"
## Problem

`load_agent_command` (`src/routines/agents/mod.rs`) collapses two distinct failure modes into a single `None`:

```rust
pub fn load_agent_command(name: &str) -> Option<AgentCommand> {
    let text = std::fs::read_to_string(agent_toml_path(name)).ok()?;   // file missing -> None
    toml::from_str(&text).ok()                                          // malformed TOML -> None
}
```

A **malformed** `~/.config/moadim/agents/<name>.toml` (valid agent name, file present, but broken TOML — e.g. a user hand-edited it; the daemon deliberately preserves user edits and never overwrites them) is indistinguishable from a **missing** file. Both yield `None`.

Downstream, `build_block` (`src/sync/routines.rs`) treats that `None` as "config not found":

```rust
None => {
    log::warn!(
        "routine sync: agent config not found for routine {:?} (agent {:?}); skipping",
        routine.id, routine.agent
    );
    None
}
```

Net effect for a routine whose agent TOML parses-fail:

- It is silently **`filter_map`-skipped** at crontab sync — gets no cron line, never fires.
- `svc_trigger` hits the same `None` branch — manual trigger spawns nothing.
- The only signal is a `log::warn!` that **misreports the cause** as "config not found" when the file is right there, just malformed. And because no logging backend is wired up (#132), even that misleading warning is discarded.

So a single typo in an agent `.toml` silently disables every routine using that agent, and the one diagnostic that exists points the user at the wrong problem (a missing file).

This is the follow-up that #139 explicitly carved out and deferred:
> Validates against the agent registry membership only, not the deeper validity of the agent's TOML (`load_agent_command` returning `None` on malformed TOML is a separate concern).

## Motivation

Fail loud, and fail *accurately*. A parse error and a missing file are different problems with different fixes; conflating them and then mislabeling the result sends debugging in the wrong direction. The fix is cheap — `toml::from_str` already returns a descriptive error; today it is thrown away by `.ok()`.

## Acceptance criteria

- [ ] `load_agent_command` distinguishes "file missing" from "file present but unparseable" (e.g. return a `Result`/enum, or log the two cases distinctly at the call site).
- [ ] The sync/trigger skip path no longer claims "config not found" for a malformed file — its diagnostic names the real cause: the agent TOML path **and** the underlying `toml` parse error.
- [ ] Extend the create/update agent validation (see #139) so that a referenced agent whose TOML is present but unparseable is rejected with `AppError::BadRequest` quoting the parse error, surfacing the problem at edit time over REST (400) and MCP rather than silently at every fire time.
- [ ] Tests cover: malformed agent TOML → distinct error/diagnostic (not the missing-file message); valid built-in (`claude`, `codex`) → unchanged behavior.

## Notes / non-goals

- Distinct from #139, which validates **registry membership** (unknown agent name) only and explicitly defers malformed-TOML handling to a separate issue.
- Making the existing silent warnings visible at all is the logging-backend gap (#132); this issue is about emitting the *correct* signal, not about where it goes.

---

> This issue was suggested by the **Suggest new GitHub issues (daemon + landing)** routine of moadim, as an unassigned suggestion. Self-assign to approve it. @tupe12334

tupe12334:
## Summary

Issue #189: `load_agent_command` collapsed two distinct failures — a *missing* config file and a *present-but-unparseable* one — into a single `None`. A malformed `~/.config/moadim/agents/<name>.toml` was therefore indistinguishable from a missing file, and downstream (`build_block` in `src/sync/routines.rs`, `svc_trigger` in `src/routines/service.rs`) it was misreported as "agent config not found" while the routine was silently skipped.

## Changes

- **`load_agent_command` now returns `Result<AgentCommand, AgentLoadError>`** (`src/routines/agents/mod.rs`). The new `AgentLoadError` enum distinguishes `Missing` (no file) from `Parse(String)` (file present, unparseable — carrying the `toml` error), with a `Display` impl that renders each distinctly.
- **Accurate skip diagnostics.** `build_block` and `svc_trigger` no longer claim "config not found" for a malformed file: the warning names the agent and the underlying parse error. The missing-file case is unchanged — still skipped and warned, just with an accurate message.
- **Reject malformed configs at edit time.** `svc_create`/`svc_update` call a new `validate_agent` helper that returns `AppError::BadRequest` (REST 400 / MCP) quoting the parse error when a referenced agent's TOML is present-but-unparseable. A *missing* config is intentionally still allowed (the agent may be registered later), preserving today's behavior.
- **Tests** cover: malformed TOML → distinct `Parse` error / skip (not the missing-file message); missing → `Missing`; valid config → unchanged `Ok` behavior; `svc_create`/`svc_update` 400 on malformed agent; `AgentLoadError` Display. Added to existing sibling `*_tests.rs` files per CONTRIBUTING.
- CHANGELOG `## [Unreleased]` entry added.

## Verification

- `cargo fmt --check` — clean
- `cargo test` — 442 passed, 0 failed
- New lines are at 100% line coverage (`cargo llvm-cov`).

Closes #189

This pull request was opened by the 'Implement my assigned issues without a PR (daemon + landing)' routine of moadim.


tupe12334:
LGTM

tupe12334:

tupe12334:
Correction to my previous reply: I could not merge this PR after all. The branch has merge conflicts with `main`, and updating the branch fails for the same reason. Resolving them requires manual judgment, and my safety rules forbid force-pushing or rewriting history, so I'm holding the merge rather than guessing.

Could you resolve the conflicts (or tell me how you'd like them resolved)? Once the branch is conflict-free I'll merge it on the next run.

tupe12334:
Update: conflicts with `main` are resolved (merged `main` into the branch via a normal merge commit — no history rewrite; the two same-named `validate_agent` functions from #189 and #139 were combined into one that performs both the unknown-agent and malformed-config checks) and all CI checks passed, so I've now squash-merged this PR and deleted the branch. Thanks for the LGTM.

PR #235, opened 2026-06-18
ci: gate CHANGELOG.md in pre-push to mirror Changelog workflow

### LINKED ISSUE #234: pre-push hook should mirror the CHANGELOG CI gate
## Problem

The `Changelog` CI workflow (`.github/workflows/changelog.yml`) fails any PR that touches `src/` or `ui/` without a `CHANGELOG.md` update (via `dangoslen/changelog-enforcer`, bypass = `skip-changelog` label).

The local `.githooks/pre-push` hook runs test-convention, fmt, clippy, and coverage — but **not** the changelog check. So a push succeeds locally and only fails later in CI, costing a round-trip.

## Proposal

Add a changelog step to `.githooks/pre-push` that mirrors the CI logic:

- Diff the range being pushed against `origin/main` (merge-base).
- If any `src/**` or `ui/**` path changed but `CHANGELOG.md` did not → fail with a helpful message.
- Provide a local bypass (`SKIP_CHANGELOG=1`) equivalent to the `skip-changelog` label.

Catches the failure before it reaches CI/CD.

tupe12334:
Closes #234.

## Problem

The `Changelog` CI workflow (`.github/workflows/changelog.yml`) fails any PR touching `src/` or `ui/` without a `CHANGELOG.md` update. The local `.githooks/pre-push` hook ran fmt, clippy, coverage, and the test-convention check — but **not** the changelog gate — so the failure only surfaced in CI after a push, costing a round-trip.

## Change

Add a 5th step to `.githooks/pre-push` mirroring the CI logic:

- Diff the pushed range against `origin/main` (merge-base).
- If any `src/**` or `ui/**` path changed but `CHANGELOG.md` did not → fail with a helpful message.
- `SKIP_CHANGELOG=1` bypasses it — the local equivalent of the `skip-changelog` PR label.
- Gracefully skips when there's no `origin/main` to diff against.

This PR itself touches only `.githooks/`, so neither the CI gate nor the new local gate require a changelog entry.


tupe12334:
LGTM

tupe12334:

PR #172, opened 2026-06-17
fix(routines): validate agent on create/update (#139)

tupe12334:
## Problem

A routine's `agent` field is resolved at runtime against `~/.config/moadim/agents/<name>.toml`, but nothing validated it on create or update. `svc_create` / `svc_update` copied `req.agent` straight onto the `Routine` with no check that the agent exists.

A typo or a name not in the registry (`available_agents()`) was accepted, persisted, and shown as a normal `enabled` routine — then failed to fire **silently** at every path:

- **Crontab sync** (`build_block`): `load_agent_command` returns `None`, the routine is `filter_map`-skipped with only a `log::warn!`, no crontab line.
- **Manual trigger** (`svc_trigger`): same `None` branch — logs a warning, spawns nothing.

With no logging backend installed (#132) those warnings are discarded. Net effect: the routine looks healthy everywhere but never runs, with no signal why.

## Fix

Validate `agent` against `available_agents()` in `svc_create` and `svc_update` (when `req.agent` is `Some`), before persisting. Unknown agents are rejected with `AppError::BadRequest` (400 over REST/MCP) naming the unknown agent and listing the valid ones — mirroring the existing `validate_cron` and slug-conflict guards. Fail loud at the point of the mistake instead of silently at fire time.

## Acceptance criteria

- [x] `svc_create` rejects an unknown `agent` with `BadRequest`, naming it and listing valid agents.
- [x] `svc_update` applies the same check when `req.agent` is `Some(...)`, before persisting.
- [x] Error surfaced over REST (400) and MCP.
- [x] Tests: create with unknown agent → 400 (nothing persisted); update to unknown agent → 400; create with a valid built-in (`claude`) → success.
- [x] No behavior change for routines referencing a valid agent.

Scope matches the issue's non-goals: validates registry membership only, not the deeper TOML validity, and does not change the existing silent-skip warnings (the logging-backend gap, #132).

`cargo fmt`, `cargo clippy --all-targets` (strict lints), and `cargo test routines::service` (16 passed) all green locally.

Closes #139

---
🤖 This pull request was opened by the *Open issues → fix PRs (owned repos + my orgs)* routine of moadim, on behalf of @tupe12334.

PR #248, opened 2026-06-18
fix(tests): repair routine lifecycle tests broken by agent validation (#172)

tupe12334:
## Why

The HTTP and MCP routine **lifecycle integration tests** have been failing on `main` since #172 added `validate_agent` on routine create/update:

- `routes::http::http_tests::router_routine_full_lifecycle`
- `routes::mcp::mcp_tests::create_get_update_trigger_delete_routine_success`

#172 added the validation and updated `service_tests`, but these two integration tests still POST/construct routines with placeholder agent names (`http-test-agent-x` / `mcp-routine-agent-x`) that aren't in the agent registry. With validation in place, creation now returns **400 / a tool error**, so both tests panic:

```
router_routine_full_lifecycle  — left: 400, right: 201
create_get_update_trigger_delete_routine_success — assertion failed: !result.is_error...
```

A red suite means `cargo test` (and therefore the `cargo llvm-cov` line-coverage step) aborts, so the pre-push **test + 100% coverage gate** can never pass on HEAD.

## What

Fix the tests rather than weaken the validation:

- Target the built-in **`claude`** agent so create/update pass `validate_agent`.
- Run each test under a fresh **`MOADIM_HOME_OVERRIDE`** temp home (the existing test seam). With no `claude.toml` seeded there, the registry falls back to the built-in defaults (so the name validates) while `load_agent_command` still returns `None` — preserving the **"trigger records, no spawn"** behavior these tests assert, and keeping the real `~/.config/moadim` untouched.

The `TempHome` guard mirrors the seam already used by `routine_storage_tests` / `cleanup_tests` / `service::mod_tests`.

## Result

No production code changes. Test-only.

```
before: 442 passed; 2 failed
after:  444 passed; 0 failed
```

`cargo fmt --check` and `cargo clippy` are clean.

---
This pull request was opened by the "Daemon + Landing auto-improve PRs" routine of moadim.


tupe12334:
LGTM

tupe12334:

PR #249, opened 2026-06-18
fix(cleanup): measure workbench retention from run finish, not trigger

### LINKED ISSUE #174: Workbench retention is measured from run start, not run finish — long/short-interval runs lose their logs immediately
## Problem

Workbench retention is measured from the **trigger timestamp** (when the run *started*), not from when the run *finished*.

`parse_workbench_name` (`src/routines/cleanup/mod.rs`) extracts the trailing `{ts}` from `{slug}-{ts}` — that `ts` is the launch time stamped in `build_routine_command` (`WB="$HOME/.moadim/workbenches/$SLUG-$TS"`, `TS="$(date +%s)"`). `is_expired(now, ts, ttl)` then reaps a finished workbench once `now - trigger_ts > ttl`.

But `effective_ttl_secs` is `min(MAX_TTL_SECS, cron_interval)` (`src/routines/cleanup/ttl.rs`), and its own doc says the intent is *"a finished run is worth keeping only until the next run is due."* Measuring age from the **start** of the run means the run's own duration is subtracted from its retention window. A run consumes its keep-window while it is still executing.

## Motivation

The window meant to preserve a finished run's logs/output (`agent.log`, cloned repos, `prompt.md`) is silently shortened — sometimes to zero:

- Routine on `*/30 * * * *` → effective TTL = 30 min. A run that starts at `:00` and takes 25 min finishes at `:25` with only ~5 min of nominal retention remaining. The next hourly sweep reaps it almost immediately after it finished.
- A run whose duration **exceeds** its effective TTL (long task on a short-interval routine, or any run on a sub-hour schedule) finishes **already past expiry** and is deleted on the very next sweep — its logs can vanish seconds after completion, before anyone reads them.

This is the retention semantics, distinct from:
- **#170** (sweep cadence is hourly while TTL can be ~60s → finished workbenches *pile up* between sweeps) — opposite direction.
- **#137** (no max-runtime watchdog for *hung* sessions) — about killing live sessions, not retaining finished ones.

The repo already assumes finish-time semantics elsewhere: a TODO item asks to *"Show a humanized retention countdown per finished run ... derived from the run's finish time and the routine's effective TTL"* — but the reaper has no finish time to derive from.

## Proposed direction

Base expiry on when the run **finished**, not when it started:

- Record a finish time per workbench — e.g. write a `.finished` marker (or touch a sentinel) when the tmux session exits, or fall back to the mtime of `agent.log` (last agent write) as a proxy.
- In `reap_dir`, compute age as `now - finish_time` instead of `now - trigger_ts`.
- Keep the existing live-session guard (`tmux_session_alive`) so running sessions are never touched, and keep `MAX_TTL_SECS` as the ceiling.

## Acceptance criteria

- [ ] A finished workbench is retained for its full `effective_ttl_secs` measured **from run completion**, regardless of how long the run took.
- [ ] A run whose duration exceeds its effective TTL is still retained for the full TTL after it finishes (not reaped on the next sweep).
- [ ] Live sessions remain protected; orphaned workbenches (routine deleted) still fall back to `MAX_TTL_SECS`.
- [ ] Unit tests cover: run shorter than TTL, run longer than TTL, and a finish time far in the past — asserting reap decisions are computed from finish time, not trigger time. Decision logic stays injectable (no real clock/tmux), per the existing `reap_dir` test design.

---
*This issue was suggested by the "Suggest new GitHub issues (daemon + landing)" routine of moadim, as an unassigned suggestion. Self-assign to approve it.*

tupe12334:
## Problem

Workbench retention was measured from the **trigger** timestamp (the `{slug}-{ts}` launch time stamped into the workbench name), not from when the run *finished*. `effective_ttl_secs` is `min(MAX_TTL_SECS, cron_interval)` — intended to keep a finished run "only until the next run is due" — so measuring age from the start means the run's own duration is subtracted from its retention window:

- A `*/30 * * * *` routine (TTL 30 min) whose run starts at `:00` and takes 25 min finishes at `:25` with ~5 min nominal retention left.
- A run whose duration **exceeds** its effective TTL finishes **already past expiry** and is reaped on the very next sweep — logs can vanish seconds after completion.

## Fix

Base the TTL reap decision on run **finish** time instead of trigger time:

- `reap_dir` takes an injected `finished_at(path, trigger_ts) -> u64` and uses it for the TTL check.
- Production resolver `agent_log_finish_time` uses the mtime of the run's `agent.log` (the last agent write) as the finish proxy, **clamped to at least `trigger_ts`** and falling back to `trigger_ts` when the log is missing/unreadable — so retention is never measured from before the run started.
- The max-runtime **watchdog** still measures elapsed wall-clock from `trigger_ts` (correct), and the live-session guard (`tmux_session_alive`) plus the `MAX_TTL_SECS` ceiling are unchanged.

One consequent improvement: a hung run that the watchdog force-kills (which writes to `agent.log`) is now retained for the full TTL after the kill rather than reaped in the same sweep — its logs survive the retention window.

## Tests

- `reap_dir_measures_ttl_from_finish_not_trigger` — same trigger age for two runs; only the long-finished one is reaped, the just-finished long run is kept.
- `agent_log_finish_time_falls_back_to_trigger_without_log` and `agent_log_finish_time_uses_log_mtime_clamped_to_trigger`.
- All existing `reap_dir` tests updated to the new signature (finish == trigger), preserving their assertions. Decision logic stays injectable — no real clock/tmux.

`cargo fmt`, `cargo clippy --all-targets`, and `cargo test cleanup` (40 tests) pass locally. Note: two pre-existing lifecycle tests (`router_routine_full_lifecycle`, `create_get_update_trigger_delete_routine_success`) fail identically on clean `main` in this sandbox (a custom test agent isn't registered) — unrelated to this change, which does not touch validation or routing.

Closes #174

---
🤖 This pull request was opened by the "Open issues → fix PRs (owned repos + my orgs)" routine of moadim.


tupe12334:
LGTM

tupe12334:

tupe12334:

Issue #141, opened 2026-06-17
Overlapping runs: a new routine session launches even when the previous run is still alive


tupe12334:
> **Suggestion** filed by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. Unassigned on purpose — self-assign to approve.

## Problem

Every cron tick and every manual trigger builds a launch command keyed by a fresh timestamp: the workbench is `~/.moadim/workbenches/{slug}-$TS` and the tmux session is `moadim-{slug}-$TS` (`build_routine_command` in `src/routines/command.rs`). Nothing checks whether the routine's **previous** run is still running before starting a new one.

So if a routine is scheduled more frequently than a run takes to finish (or it is triggered manually while a scheduled run is in flight), the daemon happily spawns a second, third, … concurrent agent session — each cloning the same repos into its own workbench and acting on them in parallel.

Consequences:
- Multiple agents racing on the same repositories (duplicate PRs/issues/commits, clobbered local clones).
- Resource pile-up (N concurrent tmux sessions + agent processes) for any slow routine on a tight schedule.
- Confusing `LOGS` view: `svc_logs` only ever surfaces the *newest* workbench, silently hiding the other in-flight runs.

Note the building block already exists: `tmux_session_alive` in `src/routines/cleanup/mod.rs` does an exact `tmux has-session -t ={session}` check — it is used to protect live workbenches from cleanup, but **not** consulted at launch time.

## Motivation

A routine is conceptually a singleton schedule. Two copies of the same task running at once is almost never what the user wants, and on shared repos it is actively harmful. This is distinct from #137 (kill a *hung* run after a max runtime): that reaps a stuck session; this prevents a healthy-but-slow run from being doubled up. The two are complementary.

## Proposed approach

Before launching, skip the new run if a session for the same routine slug is already alive. Because the session name carries a timestamp (`moadim-{slug}-$TS`), an exact-name check is not enough — match by prefix on the slug instead, e.g. early in the `;`-joined command:

```sh
tmux list-sessions -F '#S' 2>/dev/null | grep -q "^moadim-$SLUG-[0-9]\+$" && { echo "moadim: routine $SLUG already running; skipping this tick" >&2; exit 0; }
```

Design decisions to settle in the issue:
- **Default policy**: skip-if-running (recommended) vs. queue vs. allow-overlap.
- Whether it should be **configurable per routine** (e.g. an `allow_overlap: bool` on `Routine`), defaulting to skip.
[...]

PR #291, opened 2026-06-18
fix: version-stamp run.sh and refuse older-binary overwrite (#167)

### LINKED ISSUE #167: Stale daemon silently strips routine-origin disclosure when regenerating run.sh
## Summary

Generated `run.sh` files are rewritten in place by whatever daemon binary is running. A binary **older than #135 / v0.11.0** regenerates them **without** the routine-origin disclosure block — silently dropping the transparency guarantee that a newer binary had written. No version guard, no warning, no detection.

## Evidence

In `~/.config/moadim` (a git repo tracking the generated routine files):

- Committed `run.sh` (produced by a v0.11.0+ daemon) contains the `## Routine origin disclosure` section + `Routine name: ...`, via a `printf '%b\n\n%b%s\n\n...'` with the disclosure text + routine name as args.
- Working-tree `run.sh` (regenerated by the currently-running, older daemon) drops those args entirely: `printf '%b\n\n**Run date**...'` — **no disclosure**.
- This is not isolated: `git status` shows ~18 routine `run.sh` files modified the same way in one regen pass.

So every outward-facing action (GitHub PRs/issues/comments, Slack, email) produced by these routines no longer discloses that it originates from a moadim routine — the exact guarantee `MOADIM_DISCLOSURE` (`src/routines/command.rs:137`) and its test (`src/routines/mod_tests.rs:140`) are meant to enforce.

## Root cause

`run.sh` content is a pure function of the daemon binary version, regenerated in place with no idempotency/version awareness. Downgrading or running a stale binary silently rewrites generated files and removes features (here: the disclosure). The source is correct; the deployed/running binary is behind.

## Impact

- Transparency/disclosure regression on bot actions — silent, no signal to the user.
- General class of bug: any generated-file feature can be silently reverted by an older running daemon.

## Suggested fixes

- Stamp generated `run.sh` with the daemon version that wrote it; refuse to overwrite a newer-stamped file with an older binary (or at least warn).
- On startup, detect a running daemon older than the on-disk binary and require restart (ties into #141 / overlapping-run concerns).
- Add a regression guard: regeneration must never remove the disclosure block.

## Related

- #135 (added disclosure, v0.11.0)
- #152 (disclosure not reaching Codex agent — different gap, same feature)
- #151 (gitignore run.sh)

tupe12334:
## Why

Generated routine `run.sh` scripts are rewritten in place on every sync as a pure function of the running daemon's source. A binary **older than the one that last wrote a script** silently regenerated it **without** whatever the newer binary had added — most damagingly the routine-origin disclosure block (`MOADIM_DISCLOSURE`) — with no version guard, no warning, no detection. Every outward-facing action produced afterward then lost the moadim-origin disclosure. Closes #167.

## What

- `write_routine_script` now stamps every `run.sh` with a second line
  `# moadim-version: X.Y.Z` (from `CARGO_PKG_VERSION`).
- Before rewriting, it reads the existing script's stamp. If that stamp is a
  **newer** semver than the running binary, the script is **left untouched** (a
  warning is logged). The returned path still points at the preserved script, so
  the crontab line stays valid and scheduling is unaffected.
- Older-stamped or unstamped scripts (e.g. written before this feature) are
  regenerated normally — the upgrade direction is never blocked.

New helpers: `parse_version` (semver-core → comparable tuple, ignoring
`-prerelease`/`+build`) and `script_stamped_version` (reads the stamp, `None`
when the file is absent/unstamped).

### Scope note

The root cause is a *deployment* condition (a stale binary running); the in-repo
source was already correct and guarded by an existing test. This PR adds the
forward-protection the issue asks for ("stamp `run.sh` with the daemon version …
refuse to overwrite a newer-stamped file with an older binary") plus a
regression assertion that the regenerated `run.sh` still carries the disclosure
block. It does not attempt the separate startup stale-daemon detection (ties
into #141) — that is left as a follow-up.

## Verification

- `cargo fmt --check` — clean
- `cargo test` — 470 passing, incl. new `parse_version_*`,
  `script_stamped_version_*`, and three `write_routine_script_*` cases
  (stamps current version; refuses newer; overwrites older)
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` —
  100% line coverage maintained

---
This pull request was opened by the *Implement my assigned issues without a PR (daemon + landing)* routine of moadim.


tupe12334:
Blocked: superseded by upstream. `main` no longer generates per-routine `run.sh` scripts at all — crontab lines now invoke `moadim schedule trigger <id>` directly (`src/sync/routines.rs`), so the version-stamp/downgrade-guard this PR adds has no code left to attach to. Needs a manual look (likely just closing).

— PR comment commander

tupe12334:
Blocked: superseded. Issue #167 is already closed by merged PR #893 ("fix(daemon): warn when the running binary drifts from the one on disk"), already on `main`. Nothing left to merge — recommend closing as superseded.

— PR comment commander

PR #312, opened 2026-06-19
fix(openapi): derive spec version from CARGO_PKG_VERSION, not a frozen 0.1.0 (#309)

### LINKED ISSUE #309: OpenAPI spec advertises a frozen version (0.1.0) — drifted from the crate's actual 0.12.0; derive it from CARGO_PKG_VERSION
## Problem

The OpenAPI spec hardcodes its `info.version` as `0.1.0`, but the crate is currently at `0.12.0`:

- `src/openapi.rs:5` — `info(title = "Moadim Server API", version = "0.1.0", ...)`
- `Cargo.toml:12` — `version = "0.12.0"`

This spec is served live at `/docs/openapi.json` (and the Swagger UI at `/docs`, see `src/routes/http.rs:203`), so every consumer — API clients, codegen tools, Swagger UI users — sees `0.1.0` regardless of which daemon version they're actually talking to. The value has been frozen since the spec was first written and will keep drifting with every release.

Note the `GET /health` response already reports the *real* version via `CARGO_PKG_VERSION` (`src/routes/http.rs:29`), so the API self-contradicts: `/health` says `0.12.0` while `/docs/openapi.json` says `0.1.0`.

## Motivation

- The OpenAPI document is the machine-readable contract; a wrong version there silently misleads codegen/clients and breaks version-pinning workflows.
- It's a single source-of-truth drift that no test or CI step currently catches.
- `utoipa` supports deriving the version from the package automatically (e.g. omit the `version` field, or set `version = env!("CARGO_PKG_VERSION")`), so the fix removes the manual step entirely rather than swapping one hardcoded string for another.

## Acceptance criteria

- [ ] `info.version` in the served OpenAPI spec equals `CARGO_PKG_VERSION` (currently `0.12.0`) instead of a literal, so it tracks `Cargo.toml` automatically on every release.
- [ ] A test asserts the served spec's `info.version` matches `env!("CARGO_PKG_VERSION")`, so the two can't silently diverge again (fits the repo's 100% line-coverage gate).
- [ ] `/docs/openapi.json` `info.version` and the `GET /health` `version` field agree.

---
*This issue was suggested by the "Suggest new GitHub issues (daemon + landing)" routine of moadim, as a non-binding suggestion. Left unassigned — self-assign to approve. @tupe12334*

tupe12334:
## Why

The OpenAPI document hardcoded `info.version = "0.1.0"` in the `#[openapi(info(...))]` attribute of `src/openapi.rs`. The crate is actually at **0.12.0**, so the spec advertised a stale version everywhere it surfaces:

- the live REST spec served at `GET /api/v1`
- the embedded Swagger UI at `/docs`
- the committed `apis/openapi.json` snapshot

Consumers generating clients or pinning against the version saw a number that hadn't moved in many releases.

## What

Drop the `version` literal from the `info(...)` attribute. utoipa then derives `info.version` from `CARGO_PKG_VERSION` — the same Cargo metadata source it already uses to populate the MIT `license` block. Title and description stay explicit (they intentionally differ from the Cargo `name`/`description`).

`src/openapi_tests.rs::committed_spec_is_current` regenerated `apis/openapi.json`; its `info.version` is now `0.12.0` and will track every future version bump automatically.

## Checks
- `cargo test openapi` — passes (regenerated the committed snapshot)
- `cargo fmt --check` — clean
- `cargo clippy` — clean
- `CHANGELOG.md` updated under `[Unreleased] → Fixed`

Closes #309.

---


tupe12334:
LGTM

tupe12334:

PR #313, opened 2026-06-19
fix(cli): preflight foreground start for an already-running daemon (#298)

### LINKED ISSUE #298: Foreground start (`moadim -i`) has no "already running" preflight — second instance dies with an opaque `Address already in use (os error 48)`
> _Suggestion filed by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. Unassigned on purpose — self-assign to approve it._

## Problem

`run_server()` (the foreground path, `moadim -i` / `-f`) binds the listener directly with no liveness preflight:

```rust
// src/main.rs — run_server()
let listener = tokio::net::TcpListener::bind(cli::bind_addr()).await?;
```

Background mode does the opposite — it checks first and recovers gracefully:

```rust
// src/cli.rs — run_background()
if is_running() {
    let pid = read_pid_file().map(|p| format!(" (pid {p})")).unwrap_or_default();
    println!("moadim is already running{pid}; stopping it to start a fresh instance");
    crate::restart::stop_running_and_wait()?;
}
```

So starting the server in the foreground while a daemon (or any process) already holds `127.0.0.1:5784` aborts with the raw I/O error propagated through `?`:

```
Error: Address already in use (os error 48)
```

No pid, no address context, no mention of `moadim stop` / `moadim restart` / `MOADIM_BIND_ADDR`.

## Motivation

- Inconsistent DX between the two documented start modes: background self-heals, foreground dumps an opaque OS error.
- `-i` / `--interactive` is a first-class, README-documented mode — its most common failure (port already taken) gives the operator nothing actionable.
- The codebase already cares about script-friendly, predictable exits (`EXIT_NOT_RUNNING`, the status/stop/cleanup contract); a bare `os error 48` breaks that consistency.

## Acceptance criteria

- Before binding in foreground, detect an occupied port and emit a clear, actionable message naming the address, the existing pid when known, and the remedies (`moadim stop`, `moadim restart`, or set `MOADIM_BIND_ADDR`).
- Distinguish "another moadim is running" (`is_running()` true) from "port held by a foreign process" (bind fails but `is_running()` false) in the wording.
- Exit nonzero with a stable code consistent with the existing CLI exit-code convention.
- Add a test seam covering the preflight decision, mirroring the existing `cli_tests.rs` style.

## Out of scope

- Auto-restarting in foreground (background already owns that behavior); this is about a clear diagnostic + clean exit, not silently killing the running instance.

tupe12334:
## Why

Starting a second interactive server (`moadim -i` / `--interactive`) goes straight to binding the address and dies with an opaque OS error:

```
Error: Address already in use (os error 48)
```

Nothing tells the user a real daemon is already up or how to resolve it. The background path (`run_background`) already detects this case (it stops and replaces the running instance); the foreground path had no equivalent preflight. This is issue #298.

## What

- Add `ensure_not_running_for_foreground()`, called from the `Command::Foreground` dispatch in `main.rs` before binding. When a server is already reachable on the bind address it refuses with an actionable message — naming the running pid when known and pointing at `moadim stop` / `moadim restart` — instead of letting the bind fail later.
- An interactive run **refuses** rather than silently replacing the running instance (unlike `run_background`): killing the existing daemon would be a surprising side effect of `-i`.
- The launcher-spawned background child also runs with `--interactive`, but it *is* the freshly started server, so the preflight is skipped for it via the existing `MOADIM_DAEMONIZED` marker — background/restart launches are unaffected.
- The network probe (`ensure_not_running_for_foreground`) is split from the pure decision (`foreground_preflight`) so both the refuse and proceed branches are unit-testable without a live server.

## Tests

Added unit tests covering the message formatting (with/without pid), both `foreground_preflight` branches, and `ensure_not_running_for_foreground` across the not-running, server-up, and daemonized-child paths (reusing the existing `FakeServer` / `EnvGuard` harness).

## Verification

- `cargo fmt --check` ✅
- `cargo clippy --all-targets` ✅ (clean)
- `cargo test` ✅ — all new tests pass; the only failing test (`restart_tests::stop_running_and_wait_force_kills_then_succeeds_when_server_goes_down`) is unrelated: it binds the fixed default `127.0.0.1:5784` and fails in this environment because a real `moadim` daemon is already listening there, not because of this change.

Closes #298.

---


tupe12334:
LGTM

tupe12334:

PR #321, opened 2026-06-19
fix(tests): de-flake restart stop-and-wait force-kill test

tupe12334:
## Why

The pre-push hook runs `cargo fmt --check` + `cargo clippy` + `cargo llvm-cov --fail-under-lines 100`. On current stable the hook fails **intermittently** — not because of the code, but because of a flaky test.

`stop_running_and_wait_force_kills_then_succeeds_when_server_goes_down` couples three independent wall-clock timers:

- two 80ms stop-and-wait windows inside `stop_running_and_wait`, and
- a detached `stop_after(Duration::from_millis(130))` thread that flips the fake server to "down".

The success path needs the 130ms server-down event to land **inside** the post-kill wait window. Under load — notably the `cargo llvm-cov` instrumented build the hook uses — the detached timer or the poll loop drifts, the server-down event falls outside the window, `stop_running_and_wait` bails, and the `.expect(...)` panics.

Reproduced here: a plain `cargo test` run passed (477/477) while the immediately following `cargo llvm-cov --fail-under-lines 100` run failed with exit 101 on exactly this test.

## What

Tie "server down" to the **force-kill event** instead of a clock:

- The test diverts the killer through the existing `MOADIM_KILL_BIN` seam to a shim that creates a marker file (it signals no real PID).
- The fake server gains a `start_until_marker(path)` constructor and stops answering once that marker exists.

The first wait still times out with the server up; the force-kill then drops the server deterministically, so the post-kill wait always observes it stopped. No wall-clock race remains, and the assertion is unchanged.

This also removes the now-unused `stop_after` helper and the real `sleep 30` child process from this test (the shim no longer needs a live PID to kill).

## Verification

- `cargo fmt --check` ✅
- `cargo clippy --all-targets` ✅ (clean under the repo's `lints.clippy.all = "deny"`)
- restart suite: **15/15** green runs
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'`: exit **0** on two consecutive runs (previously exit 101); line coverage stays 100%

Scope is one test module (`src/restart_tests.rs`); no production code changed.

---

This pull request was opened by the **Daemon + Landing auto-improve PRs** routine of moadim. @tupe12334


tupe12334:
LGTM

tupe12334:

PR #299, opened 2026-06-18
fix(cli): `stop --json` reports effective bind address, not hardcoded default

tupe12334:
## Why

`status --json` and `stop --json` are designed to emit the **same** object shape (`{"running":bool,"pid":N|null,"address":…}`) so callers can parse both uniformly. But the `address` field had silently drifted apart:

- `status_json()` calls `bind_addr()`, which honors the `MOADIM_BIND_ADDR` override.
- `stop_json()` hardcoded the `BIND_ADDR` constant (`127.0.0.1:5784`).

So whenever a custom bind address was configured, `status --json` reported the real address while `stop --json` kept emitting the default — exactly the kind of drift the shared shape was meant to prevent. (The recent "`moadim status` reports the effective bind address" fix never reached `stop`.)

## What

- `stop_json()` now uses `bind_addr()`, matching `status_json()`.
- Added `stop_json_address_reflects_bind_override()`, mirroring the existing `status_json_address_reflects_bind_override()` test, so the invariant is now pinned for both commands.
- Corrected the `stop_json` doc comment, which wrongly promised the hardcoded `BIND_ADDR`.
- CHANGELOG `[Unreleased]` entry added.

## Checks

- `cargo fmt --check` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --bin moadim cli` — 44 passed, incl. the new test

Small, focused, no behavior change beyond the corrected field.

---
This pull request was opened by the **Daemon + Landing auto-improve PRs** routine of moadim.

Issue #284, opened 2026-06-18
Enrich `moadim status --json` with server uptime & version from /health (single call returns liveness + age)


tupe12334:
## Problem

`moadim status --json` returns only local-process facts:

```json
{"running": true, "pid": 1234, "address": "127.0.0.1:8080"}
```

(`status_json` in `src/cli.rs:341`). But the server's `GET /health` already exposes richer liveness — `uptime_secs` and `version` — via `HealthResponse` (`src/routes/http.rs:19-30`). A script that wants both "is it up?" and "how long / which version?" must make two calls (parse `status --json`, then separately hit `/health`), and the `--json` contract silently omits data the daemon already knows.

## Motivation

- A single `moadim status --json` call should answer running-state **and** age/version, not just the local PID — the operator shouldn't shell out to `curl /health | jq` to learn uptime.
- This is the enrichment called for in `TODO.md`: *"Enrich `moadim status --json` with the server's liveness details from `GET /health` (e.g. `uptime_secs`) so a single call returns running-state + age, not just the local PID."*
- Complements the existing `--json` contract work (the `address` field, `status`/`stop`/`cleanup` shape consistency) without duplicating it — this adds *server-sourced* fields, where those issues add *local* ones.

## Proposed shape

When the probe succeeds, fold the `/health` payload into the object:

```json
{"running": true, "pid": 1234, "address": "127.0.0.1:8080", "uptime_secs": 8123, "version": "x.y.z"}
```

When no server answers, `uptime_secs`/`version` are `null` (or omitted), and `running:false`/`pid` behave exactly as today. The existing exit-code contract (0 up / 3 down) is unchanged.

## Acceptance criteria

- [ ] `status --json` against a running daemon includes `uptime_secs` and `version`, sourced from a `GET /health` call (no second user call needed).
- [ ] `status --json` against a down daemon still emits a valid object with `running:false` and no crash (server fields `null`/absent).
- [ ] Exit codes and the human-readable (non-`--json`) `status` output are unchanged.
- [ ] A `cli_tests` case asserts the enriched keys are present when a listener answers `/health`, and absent/null when it does not.
- [ ] Help text / README documenting the `--json` shape is updated to list the new fields.

---
*This issue was suggested by the "Suggest new GitHub issues (daemon + landing)" routine of moadim, as an unassigned suggestion. Self-assign to approve it. @tupe12334*

Issue #245, opened 2026-06-18
cleanup --json omits the `address` field that status/stop --json emit — inconsistent machine-readable contract


tupe12334:
## Problem

The `--json` flavors of the CLI status commands are documented as a uniform machine-readable contract, but `cleanup --json` is missing the `address` field that `status --json` and `stop --json` both emit.

In `src/cli.rs`:

- `status_json(running, pid)` → `{"running":bool,"pid":N|null,"address":…}`
- `stop_json(running, pid)` → `{"running":bool,"pid":N|null,"address":…}`
- `cleanup_json(removed, running)` → `{"running":bool,"removed":N}`  ← no `address`

So a script that calls `moadim cleanup --json` cannot learn which endpoint the CLI talked to, whereas `status`/`stop` both surface it. The shapes silently diverge.

## Motivation

`--json` exists so scripts can branch without parsing human text. Every `--json` command surfacing the bound `address` makes the contract uniform: one `jq '.address'` works across `status`, `stop`, and `cleanup`. This matches the existing TODO.md item ("Have `moadim cleanup --json` include the bound `address` field too").

## Acceptance criteria

- [ ] `cleanup_json` includes `"address"` (value from `bind_addr()`), matching `status_json`/`stop_json`.
- [ ] `cleanup --json` emits `{"running":bool,"removed":N,"address":…}` in both the running and not-running branches.
- [ ] The doc comment on `cleanup_json` and the `print_help` `--json` note are updated to reflect the new shape.
- [ ] A `cli_tests` assertion locks the contract: `address` is present in `status --json`, `stop --json`, and `cleanup --json` (and ideally that the three share the same `address` value), so the shapes can't drift again.

---
*This issue was suggested by the "Suggest new GitHub issues (daemon + landing)" routine of moadim, as an unassigned suggestion. Self-assign to approve it.*

PR #325, opened 2026-06-19
test(cli): guard that `status --json` and `stop --json` emit identical key sets (#318)

### LINKED ISSUE #318: cli_tests: no regression guard that `status --json` and `stop --json` emit identical key sets — the documented "matching exactly" contract is unenforced
## Problem

`moadim status --json` and `moadim stop --json` are documented to emit the **same** machine-readable shape. The source says so explicitly:

- `src/cli.rs:232` — stop emits `{"running":bool,"pid":N|null,"address":…}`, "matching `status --json`'s shape exactly"
- `src/cli.rs:267` — same wording again

But the two payloads are built by **two independent functions** — `status_json(...)` (`src/cli.rs:341`) and `stop_json(...)` (`src/cli.rs:272`) — and nothing pins them together. The existing `cli_tests.rs` only checks each function in isolation (`status_json_reports_running_pid_and_address`, the stop equivalent, etc.). No test asserts that the two emit the **same set of object keys**.

So the "matching exactly" contract is a comment, not a guarantee. The moment one side gains a field and the other doesn't, the docs silently become false and `--json` consumers that switch between `status` and `stop` break.

## Motivation

This is a live risk, not hypothetical:

- #284 wants to enrich `status --json` with `uptime_secs`/version from `/health`. Adding that to `status_json` and forgetting `stop_json` would break the documented parity — with no test to catch it.
- The `--json` contract has already drifted once across commands (#245: `cleanup --json` is missing the `address` field that `status`/`stop` emit), proving the family of `--json` outputs drifts apart in practice.
- TODO.md explicitly calls for this guard: *"Add a `cli_tests` assertion that `status --json` and `stop --json` produce the SAME set of object keys, so the two shapes can't silently drift apart again as fields are added."*

A cheap, deterministic, pure-function test (no process spawn, no network) closes the gap permanently.

## Acceptance criteria

- A new test in `src/cli_tests.rs` parses the output of `status_json(...)` and `stop_json(...)` for the same inputs and asserts their **top-level key sets are identical** (e.g. collect `serde_json::Value::as_object().keys()` into a `BTreeSet` and `assert_eq!`).
- Covers at least both branches: a running daemon with a known PID, and a down/unknown daemon (`pid == null`), since shape must match in both states.
- The test fails if a field is added to one function but not the other (verified by a quick local spike).
- Scope is the regression guard only — this issue does **not** add `address` to `cleanup --json` (that is #245) and does not change any output shape.

---
*This issue was suggested by the "Suggest new GitHub issues (daemon + landing)" routine of moadim, as a suggestion for @tupe12334 to review. It is intentionally left unassigned — self-assign to approve.*

tupe12334:
## Why

`moadim status --json` and `moadim stop --json` are documented to emit the **same** machine-readable object shape so `--json` consumers can switch between them and parse either uniformly:

- `src/cli.rs:232`, `:267`, `:339` — both say the shape matches *exactly* (`{"running":bool,"pid":N|null,"address":…}`).

But the two payloads are built by **two independent functions** — `status_json` (`src/cli.rs:341`) and `stop_json` (`src/cli.rs:272`) — and nothing pins them together. The existing `cli_tests` only check each function in isolation, so the "matching exactly" contract is a comment, not a guarantee. The moment one side gains a field and the other doesn't, the docs silently become false and consumers break.

This is a **live risk**, not hypothetical:
- #284 wants to enrich `status --json` with `uptime_secs`/version — adding it to `status_json` and forgetting `stop_json` would break parity with no test to catch it.
- The `--json` family has already drifted once (#245: `cleanup --json` missing the `address` field).
- `TODO.md` explicitly calls for this guard.

## What

Adds one pure-function `cli_tests` regression guard (`status_and_stop_json_emit_identical_key_sets`) that parses both outputs for the same inputs and asserts their **top-level key sets are identical** — collected into a `BTreeSet` for an order-independent compare — across both the running and the down/null-pid branches. Plus a small `json_key_set` helper and a CHANGELOG entry.

Test-only, no behavior change. No process spawn, no network. It compares **keys, not values**, so it's orthogonal to #299 (which fixes the `address` *value* in `stop_json`).

## Validation

- `cargo test --bin moadim cli_tests::` → 43 pass (was 42)
- `cargo fmt --check` → clean
- `cargo clippy --bin moadim --all-targets` → clean

Closes #318.

---


tupe12334:
LGTM

tupe12334:

PR #330, opened 2026-06-19
fix(sync): crontab marker collision wipes the routines block (#324)

### LINKED ISSUE #324: Crontab marker collision: `# BEGIN MOADIM` substring-matches `# BEGIN MOADIM-ROUTINES`, so cron-jobs sync can silently wipe the routines block
## Problem

`replace_block_with` (`src/sync/mod.rs:254`) locates the managed crontab block with **substring** search:

```rust
let begin_pos = crontab.find(begin_marker); // begin_marker = "# BEGIN MOADIM"
let end_pos   = crontab.find(end_marker);   // end_marker   = "# END MOADIM"
```

But the two managed blocks use markers where one is a **prefix of the other**:

- Cron-jobs block (`src/sync/mod.rs:40`): `# BEGIN MOADIM` / `# END MOADIM`
- Routines block (`src/sync/routines.rs:35`): `# BEGIN MOADIM-ROUTINES` / `# END MOADIM-ROUTINES`

Because `"# BEGIN MOADIM"` is a prefix of `"# BEGIN MOADIM-ROUTINES"`, `crontab.find("# BEGIN MOADIM")` matches the **routines** marker line too. Verified:

```
crontab = "# BEGIN MOADIM-ROUTINES\n…# moadim-routine:abc\n# END MOADIM-ROUTINES\n"
find("# BEGIN MOADIM") -> 0      # matches the ROUTINES begin line
find("# END MOADIM")   -> 75     # matches the ROUTINES end line
# begin < end  => cron-jobs sync REPLACES the routines block
```

So when `sync_to_crontab` (cron-jobs, `src/sync/mod.rs:367`) runs and the **routines block precedes the handler block, or the handler block is absent** (common case: the user has routines but no managed cron jobs, then creates/updates/deletes a cron job — see `src/cron_jobs.rs:273,307,321`), the cron-jobs sync overwrites or corrupts the entire `# BEGIN MOADIM-ROUTINES … # END MOADIM-ROUTINES` block. Every routine crontab line is silently dropped and the routines stop firing — with no error.

The collision is **one-directional**: the routines marker `# BEGIN MOADIM-ROUTINES` is not a superstring of `# BEGIN MOADIM`, so `sync_routines_to_crontab` does not clobber the handler block. Only cron-jobs sync clobbers the routines block.

Note the writer and the parser already disagree: `parse_block` (`src/sync/mod.rs:343`) matches markers with **exact line equality** (`trimmed == BLOCK_BEGIN`), which would *not* collide — but `replace_block_with`, the function that actually rewrites the crontab, uses substring `find`.

## Motivation

Routines are the product's primary feature. A user who has routines and then touches a cron job can have all their scheduled agent runs silently removed from the OS crontab — a hard-to-diagnose data-loss/silent-no-op bug, in the same "silently never fires" family as #183, #233, #224.

## Acceptance criteria

- [ ] `replace_block_with` matches block markers on a **whole-line** basis (e.g. iterate `crontab.lines()` and compare `line.trim() == begin_marker`/`end_marker`, mirroring `parse_block`), so `# BEGIN MOADIM` never matches the `# BEGIN MOADIM-ROUTINES` line (and vice versa).
- [ ] A regression test: a crontab containing **only** the routines block, after a cron-jobs `sync_to_crontab`, still contains the intact, unmodified routines block.
- [ ] A regression test for the reverse ordering (routines block before handler block): both blocks survive independently after each sync.
- [ ] Existing block insert/replace/malformed-block behavior is preserved (the current `replace_block_with` tests still pass).
- [ ] Consider hardening the markers themselves (e.g. a trailing sentinel) so future prefix collisions are impossible, but the whole-line fix is sufficient.

---

tupe12334:
## Why

`replace_block_with` (in `src/sync/mod.rs`) located the moadim-owned crontab block with raw substring search — `str::find(begin_marker)` / `str::find(end_marker)`.

The cron-jobs block uses the markers `# BEGIN MOADIM` / `# END MOADIM`. The routines block uses `# BEGIN MOADIM-ROUTINES` / `# END MOADIM-ROUTINES`. The cron-jobs markers are **prefixes** of the routines markers, so a substring search for `# BEGIN MOADIM` happily matches the start of `# BEGIN MOADIM-ROUTINES`.

**Impact:** when a crontab held only the routines block (no cron-jobs block yet), a cron-jobs sync would resolve its delimiters to the *routines* block and silently overwrite — wipe — every scheduled routine line. Block ordering could trigger the symmetric corruption too. (Closes #324.)

## What

- Add `find_marker_line`, which matches a delimiter only when a whole line's trimmed content **equals** the marker (not a substring), and route the block replacement through it.
- The more-specific `# … MOADIM-ROUTINES` markers no longer collide with the `# … MOADIM` markers.

## Tests

New regression tests in `cron_sync_tests.rs`:
- a routines-only crontab survives a cron-jobs replace (block is appended, not substituted);
- an exact-marker replace among both blocks edits only the cron-jobs block;
- whitespace-tolerant whole-line matching.

`cargo fmt --check`, `cargo clippy`, `typos`, and the full `sync::` test suite pass locally. (The one unrelated failure seen in a full `cargo test` run — `restart::…force_kills…` — is a known pre-existing flake, passes on rerun, and is addressed separately by #321.)

---


tupe12334:
Blocked: superseded by upstream. Only one crontab block exists now (`# BEGIN/END MOADIM-ROUTINES`); no separate marker prefix exists elsewhere to collide with. The prefix-collision bug is architecturally impossible in the current design. Needs a manual look (likely just closing).

— PR comment commander

tupe12334:
Blocked: moot. The prefix-collision bug (`# BEGIN MOADIM` matching `# BEGIN MOADIM-ROUTINES`) required a second, shorter `# BEGIN MOADIM` marker from the cron-jobs block. That entire feature was removed on `main` by commit c7ad088 (#842) — only the ROUTINES marker exists now, so the collision this PR fixes can no longer happen. Recommend closing.

— PR comment commander

PR #338, opened 2026-06-19
fix(openapi): make startup spec write idempotent & installed-binary-safe (#319)

### LINKED ISSUE #319: Server rewrites apis/openapi.json to the compile-time source path on every startup — redundant in dev, a warning-only no-op in installed binaries
## Problem

`run_with_listener_until` writes the OpenAPI spec to disk on **every** server startup, unconditionally:

```rust
// src/routes/http.rs:232
write_openapi_spec(std::path::Path::new(concat!(
    env!("CARGO_MANIFEST_DIR"),
    "/apis/openapi.json"
)));
```

`env!("CARGO_MANIFEST_DIR")` is resolved **at compile time**, so the absolute path is baked into the binary:

- **In-tree dev runs:** the path points at the working copy, so the daemon silently mutates a git-tracked file (`apis/openapi.json`) as a side effect of merely *starting* the server — surprising, and a needless write on every boot/restart.
- **Installed binaries (`cargo install moadim`, OS service):** the baked path is the build machine's source tree — the cargo registry cache (`~/.cargo/registry/src/.../moadim-<ver>/apis/openapi.json`) or a path that no longer exists. The write is pointless there and, when it fails, fires `log::warn!("could not write openapi spec: …")` on **every** startup — pure log noise (compounding the unbounded-`daemon.log` growth tracked in #316).

The runtime write is also **redundant**: `openapi_tests::committed_spec_is_current` already regenerates and commits `apis/openapi.json` whenever the path decorators drift, and the test fails until it's committed. The spec is therefore kept current by the test gate at dev time; the startup write adds nothing but a side effect.

## Motivation

- A released binary should not attempt to write into its own (build-time) source tree on startup, nor emit a recurring warning for a write that can never succeed or matter on the user's machine.
- Starting the server should be free of surprise filesystem mutations to tracked files.
- Less startup log noise; one fewer best-effort failure path on the hot startup path.

## Acceptance criteria

- [ ] The server no longer writes `apis/openapi.json` on startup in release/installed builds (no spurious `could not write openapi spec` warning when the compile-time path is absent or read-only).
- [ ] The committed `apis/openapi.json` stays current through the existing test path (`committed_spec_is_current`) — i.e. spec freshness no longer depends on the daemon being run. Options: drop the startup call entirely and rely on the test, or gate it behind `#[cfg(debug_assertions)]` / a dev-only flag so it never runs in release.
- [ ] If any dev-only write is retained, it is documented as such (and does not warn when the path is missing).
- [ ] Existing tests `write_openapi_spec_writes_json_to_path` / `write_openapi_spec_logs_on_write_failure` still pass (the helper itself can remain; only its unconditional invocation at startup changes).

---
_This issue was filed as a suggestion by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. It is unassigned — approve it by self-assigning._

tupe12334:
## Why

On every startup the server writes the generated OpenAPI spec to
`CARGO_MANIFEST_DIR/apis/openapi.json` (`write_openapi_spec` in
`src/routes/http.rs`). That path is the **build machine's** source tree, which
produces two papercuts (issue #319):

- **Installed binaries:** the source tree is absent at runtime, so the write
  fails and logs `could not write openapi spec: ...` on *every* start — a
  warning about a file the operator can never have.
- **Dev checkouts:** the spec rarely changes between runs, yet the unconditional
  write bumped the committed file's mtime (and surfaced as a spurious
  working-tree change) on each start.

## What

`write_openapi_spec` now:

1. Skips silently (`debug`, not `warn`) when the parent directory doesn't exist
   — the expected installed-binary case, not an error.
2. When the parent dir exists, rewrites the spec **only if its contents
   changed**, so dev startups stop churning the committed file's mtime.

## Tests

- New `write_openapi_spec_skips_when_parent_dir_absent` — no write, no panic, no
  half-built tree.
- New `write_openapi_spec_skips_rewrite_when_unchanged` — second identical call
  leaves mtime untouched.
- Existing `write_openapi_spec_logs_on_write_failure` updated to force a real
  write error with the parent dir present (target path is a directory), keeping
  the `warn` branch covered.

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the spec
writer tests pass locally. CHANGELOG `[Unreleased] → Fixed` updated.

Closes #319.

---


tupe12334:
Blocked: superseded by upstream. `src/openapi.rs` (main) no longer writes the spec to disk at startup at all — `apis/openapi.json` is now a committed static file, kept in sync via a drift-detecting test (`openapi_tests.rs`) rather than a runtime write. The idempotent-write concern is architecturally moot. Needs a manual look (likely just closing).

— PR comment commander

tupe12334:
Blocked: superseded. Issue #319 is already addressed on `main` via commit 163ff67 ("fix(server): skip openapi.json write when the target dir doesn't exist (#1005)") — `write_openapi_spec` already skips the write when the parent dir is missing. Nothing left to merge — recommend closing as superseded.

— PR comment commander

Issue #224, opened 2026-06-18
Empty/whitespace-only routine prompt is accepted on create/update → routine fires forever with no task


tupe12334:
## Problem

`svc_create` and `svc_update` (`src/routines/service.rs`) validate **only** the cron schedule (`validate_cron`) and slug uniqueness. Neither checks that `prompt` is non-empty. An empty or whitespace-only `prompt` is accepted, persisted, and synced to the OS crontab like any other routine.

When that routine fires, `build_routine_command`/`compose_prompt` (`src/routines/command.rs`) still produce a `prompt.md` — just the moadim preamble + repo list with no task body. The launch path's only guard is "source prompt file missing → abort" (the `cp ... || exit 1` line); an empty-but-present prompt sails through. The agent launches on schedule with no task and exits (or worse, improvises), repeating every cron tick — silently burning scheduled runs and the user's agent/API budget for nothing.

This is the same "accepted on create, then silently misbehaves at run time" footgun already filed for the `agent` field (#139), 6-field cron (#183), slug collisions (#188), and malformed agent TOML (#189). The prompt — the one field that defines what the routine actually *does* — has no such guard. It matters most here because routines are commonly created over MCP by an AI agent, which can easily omit or blank the prompt.

`UpdateRoutineRequest.prompt` has the same hole: `if let Some(prompt) = req.prompt { routine.prompt = prompt; }` happily sets an existing routine's prompt to `""`.

## Motivation

- A blank-prompt routine is never useful — it is always a mistake — yet it fires indefinitely with no signal to the user.
- Wastes scheduled agent invocations and API tokens on task-less sessions.
- Consistent with the project's existing validation-hardening direction (#139, #183, #188, #189): reject bad input at the API boundary instead of silently misfiring later.

## Acceptance criteria

- [ ] `svc_create` rejects a request whose `prompt` is empty or whitespace-only with `AppError` mapping to HTTP `400` (mirroring `validate_cron`'s 400 path), before any persistence/crontab sync.
- [ ] `svc_update` rejects a request that would set `prompt` to empty/whitespace-only with `400`; a `None` prompt (field omitted) still leaves the existing prompt untouched.
- [ ] The 400 body/message names the offending field (e.g. `"prompt must not be empty"`), matching the style of existing validation errors.
- [ ] Unit tests cover: create with empty `""`, create with whitespace `"   \n"`, update clearing prompt to empty, and the happy path (non-empty prompt still succeeds).
[...]

PR #269, opened 2026-06-18
fix(cron-jobs): validate handler on create/update (#252)

tupe12334:
## Why

The cron-job `handler` field is never validated on create or update. It is later interpolated **verbatim** into a crontab line by the sync layer:

```
<schedule> <handlers_dir>/<handler> # moadim:<id>
```

(`src/sync/mod.rs::format_crontab_line`). Two concrete problems follow (issue #252):

- An **empty / whitespace-only** handler writes a broken, dangling crontab entry.
- A handler containing a **newline injects additional, attacker-controlled crontab lines** — a classic CRLF-style injection: `"evil # moadim:x\n* * * * * /bin/sh ..."` would smuggle a second schedule into the user's crontab.

`schedule` is already guarded by `validate_cron`; `handler` had no equivalent gate.

## What

- Add `validate_handler`, which rejects empty/whitespace-only values and any value containing control characters (covers `\n`, `\r`, `\t`).
- Call it from `svc_create` and `svc_update`, alongside the existing `validate_cron` check and before any store mutation.
- Add unit tests for `validate_handler` (accept + both reject paths) and service-layer tests that `svc_create` / `svc_update` return `BadRequest` for a bad handler.

Focused, test-only + a small validation helper — no behavior change for valid handlers.

## Verify

```
cargo fmt --check      # clean
cargo clippy --all-targets --all-features   # clean
cargo test cron_jobs   # 46 passed, incl. the new accept/reject tests
```

---
This pull request was opened by the "Daemon + Landing auto-improve PRs" routine of moadim.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #409, opened 2026-06-19
fix(cron): reject blank handler on cron job create/update

tupe12334:
## Why

`svc_create` and `svc_update` in `src/cron_jobs.rs` validated the cron
schedule but accepted **any** handler string — including `""` or
whitespace-only. Such a job is happily persisted to disk, yet at fire
time `svc_trigger` joins the handler onto the handlers directory and
finds nothing executable, so the job silently does nothing. There is no
feedback to the API/MCP caller that the job is dead on arrival.

The routines module already guards against exactly this class of bug:
`reject_blank` in `src/routines/service.rs` rejects empty `title`/`prompt`
before persisting (issues #224/#226). Cron jobs lacked the equivalent.

## What

- Add `validate_handler` to `src/cron_jobs.rs`, rejecting a blank /
  whitespace-only handler with `400 BadRequest` (same shape as the
  existing `validate_cron`).
- Call it in `svc_create` (after `validate_cron`) and in `svc_update`
  (when a handler is supplied). Both the REST and MCP paths go through
  these, so both are covered.
- Tests: reject empty and whitespace-only handler on create (store stays
  empty); reject blank handler on update and assert the existing handler
  is left untouched.

## Checks

- `cargo fmt --check` — clean
- `cargo clippy --all-targets` — clean
- `cargo test cron_job` — all pass, including the two new tests

Both branches of `validate_handler` are exercised by tests, keeping the
100% coverage gate satisfied.

---

### DIFF
--- a/src/cron_jobs.rs
+++ b/src/cron_jobs.rs
@@ -184,6 +184,21 @@ pub(crate) fn validate_cron(expr: &str) -> Result<(), AppError> {
     Ok(())
 }
 
+/// Reject a blank handler identifier.
+///
+/// The handler names the script under the handlers directory to spawn when the
+/// schedule fires (`svc_trigger`). A blank or whitespace-only value would be
+/// persisted as a job that can never resolve to an executable, so it is caught
+/// here before anything is written to disk.
+fn validate_handler(handler: &str) -> Result<(), AppError> {
+    if handler.trim().is_empty() {
+        return Err(AppError::BadRequest(
+            "handler must not be empty".to_string(),
+        ));
+    }
+    Ok(())
+}
+
 /// Request body for creating a new cron job.
 #[derive(Deserialize, JsonSchema, utoipa::ToSchema)]
 pub struct CreateRequest {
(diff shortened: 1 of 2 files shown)


tupe12334:
Blocked: superseded by upstream. Same underlying concern as #269 (free-form cron `handler` field) — already moot since `handler` was replaced by a whitelist-validated `agent` field. Needs a manual look (likely just closing).

— PR comment commander

tupe12334:
Blocked: moot. This PR validates handlers in `src/cron_jobs.rs`, but that entire module (cron-jobs feature) was removed on `main` by commit c7ad088 ("feat!: remove cron jobs, focus on AI-agent routines (#842)"). There is nothing left to validate — recommend closing.

— PR comment commander

tupe12334:
Blocked: this branch conflicts with `main` because the feature it targets is gone — `main` removed the whole cron-jobs REST feature in `c7ad088` "feat!: remove cron jobs, focus on AI-agent routines (#842)", including `src/cron_jobs.rs`/`src/cron_jobs_tests.rs`. There's nothing left to apply this fix to. Recommend closing as obsolete.

PR #424, opened 2026-06-19
docs(readme): document the MOADIM_BIND_ADDR override (#417)

### LINKED ISSUE #417: docs: README never documents the MOADIM_BIND_ADDR env var — the address override is source-only, and the hardcoded 127.0.0.1:5784 examples mislead anyone who sets it
## Problem

`MOADIM_BIND_ADDR` is a fully-supported, first-class environment variable: `cli::bind_addr()` (`src/cli.rs:17-20`) reads it and falls back to the `127.0.0.1:5784` default, and it drives **both** the server bind (`main.rs:88`) and every client-side CLI probe (`status` / `stop` / `cleanup`, the UI URL hint, and the `--json` `address` field).

Yet it is **never mentioned in the README**. The only reference anywhere in the repo's docs is a line in `CHANGELOG.md`. Throughout the README the address `127.0.0.1:5784` is hardcoded in prose, the CLI table, and the scripting/`--json` sections, with no hint that it is configurable.

Consequences:
- A user who wants to move the daemon off the default port (conflict on `5784`, run multiple instances, bind a non-loopback interface in a trusted setup) has no way to discover the knob short of reading the source or the changelog.
- The scripting examples (`moadim status --json | jq ...`) silently assume the default address, so they break for anyone who has set `MOADIM_BIND_ADDR` — without explaining why.

This is purely a documentation gap. It is **not** a duplicate of #253 (which asks for a *security guard/warning* when binding to a non-loopback interface) nor #385 (OpenAPI `servers` URL): this issue only asks that the existing, supported variable be documented.

## Motivation

`MOADIM_BIND_ADDR` is the single override that changes where the whole system listens and talks. Leaving it undocumented makes a supported feature effectively invisible and makes the README's hardcoded addresses misleading. Documenting it is low-effort, high-clarity, and removes a "had to read the source" papercut (cf. #16).

## Acceptance criteria

- [ ] README documents `MOADIM_BIND_ADDR`: its purpose (server bind address **and** the address the CLI probes), its default (`127.0.0.1:5784`), and the accepted form (`host:port`).
- [ ] The Running / CLI section notes that `status` / `stop` / `cleanup` / the UI URL honor the same variable, so client and server stay in agreement.
- [ ] At least one scripting example or a short note clarifies that the documented `127.0.0.1:5784` strings reflect the default and shift with `MOADIM_BIND_ADDR`.
- [ ] A brief safety note that binding to a non-loopback address exposes the unauthenticated API, cross-referencing #253 (no behavior change is requested here — documentation only).

---
*This issue was suggested (unassigned) by the "Suggest new GitHub issues (daemon + landing)" routine of moadim, as a candidate for the maintainer to approve by self-assigning. It changes no code.*

tupe12334:
## Why

The server resolves its bind/probe address from the `MOADIM_BIND_ADDR` environment variable, falling back to `127.0.0.1:5784` (`src/cli.rs`):

```rust
const BIND_ADDR_ENV: &str = "MOADIM_BIND_ADDR";
pub fn bind_addr() -> String {
    std::env::var(BIND_ADDR_ENV).unwrap_or_else(|_| BIND_ADDR.to_string())
}
```

But the README never documents it — the override was source-only, and every example (plus the `--json` payloads) hardcodes `127.0.0.1:5784`, which misleads anyone who relocates the server. Because the override changes both the bind **and** the client probe target, a client invoked without the same value keeps looking at the default port and wrongly reports the relocated server as not running.

## What

- Adds a **Bind address** subsection under *Running* documenting `MOADIM_BIND_ADDR`.
- Shows the server + client usage (both must share the value) and the export-it-to-persist tip.
- Notes that the `127.0.0.1:5784` addresses elsewhere in the docs/JSON are the default and follow `MOADIM_BIND_ADDR` when set.

Closes #417.

Docs-only change — no `src/`/`ui/` touched, so the changelog gate is exempt. `cargo fmt --check` and `typos` are clean.

---
This pull request was opened by the *Daemon + Landing auto-improve PRs* routine of moadim. @tupe12334

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,6 +315,26 @@ Starts on `http://127.0.0.1:5784`. On startup the server:
 2. Reads your crontab and applies any changes made to the moadim block while the server was stopped.
 3. Writes all enabled managed jobs back into the crontab block.
 
+### Bind address
+
+The server binds to `127.0.0.1:5784` by default — the same address every client command
+(`status`, `stop`, `cleanup`, …) probes. Override it with the `MOADIM_BIND_ADDR` environment
+variable, set identically for the server and any client you run against it:
+
+```sh
+# Bind the server to a different port (or interface)…
+MOADIM_BIND_ADDR=127.0.0.1:7000 moadim
+
+# …and point client commands at the same address.
+MOADIM_BIND_ADDR=127.0.0.1:7000 moadim status
+```
+
+Because the override changes both the bind and the probe target, a client started without it
+keeps looking at the default `127.0.0.1:5784` and will report the relocated server as not running.
+Export the variable in your shell profile to make the change stick across commands. All the
+`127.0.0.1:5784` addresses shown above and in the `--json` payloads reflect the default; they
+follow `MOADIM_BIND_ADDR` when it is set.
+
 ## MCP usage
 
 > _This is where the loop closes: your agent reads, schedules, and re-fires its own jobs. Loop engineering with a daemon in the middle._


tupe12334:
LGTM

tupe12334:
Merged — thanks for the approval.

Issue #445, opened 2026-06-19
load_agent_command collapses every I/O error into AgentLoadError::Missing — an existing-but-unreadable agent config is mislabeled "config not found" and the routine is silently dropped (green dot, never fires)


tupe12334:
## Problem

`load_agent_command` resolves an agent's `<name>.toml` like this (`src/routines/agents/mod.rs`):

```rust
let text =
    std::fs::read_to_string(agent_toml_path(name)).map_err(|_| AgentLoadError::Missing)?;
```

The `.map_err(|_| AgentLoadError::Missing)` discards the underlying `std::io::Error` and collapses **every** failure kind into `Missing` — `PermissionDenied`, `IsADirectory`, and other real I/O errors all become "config not found", even when the file demonstrably exists on disk.

This is a third, distinct failure class beyond the two already handled:

- #189 (done) split `Parse` (present-but-malformed TOML) out of `Missing`.
- #301 (open) is about `agent_registered` reflecting only `exists()`, not parseability, in the health dot.
- This issue is upstream of both: an *existing-but-unreadable* config (wrong perms, replaced by a directory, transient I/O error) is mislabeled `Missing` regardless of whether the file is there.

Concrete consequences of the misclassification:

1. **Misleading log on crontab sync** (`src/sync/routines.rs:97`): the skip warning prints `cannot load agent ... (agent config not found); skipping` for a routine whose agent config file actually exists — so the operator greps for a missing file that is right there with `chmod 000`.
2. **Silent no-op routine** combined with `agent_registered = agent_toml_path(...).exists()` (`src/routines/model.rs:154`): `exists()` is `true` (green "registered" dot), yet the routine is dropped from the generated crontab block and never fires. Healthy-looking, never runs — exactly the silent-failure class this repo keeps closing (#383, #301, #322, #189).
3. **`validate_agent`** (`src/routines/service.rs:75`) intentionally treats `Missing` as OK ("may be created later"). An unreadable-but-present config sails through create/update validation under that allowance, when it is really a broken config that should surface.

## Motivation

The whole agent-config error surface was reworked (#189) precisely so the daemon stops conflating distinct causes and reports the real one. Swallowing all `io::Error` kinds into `Missing` reintroduces that conflation for the I/O/permission case, producing a factually wrong "not found" diagnostic and a green-dot routine that never runs.

## Acceptance criteria

[...]

PR #446, opened 2026-06-19
fix(agents): distinguish unreadable agent config from missing (#445)

tupe12334:
## Why

`load_agent_command` mapped **every** `read_to_string` error to `AgentLoadError::Missing`:

```rust
std::fs::read_to_string(agent_toml_path(name)).map_err(|_| AgentLoadError::Missing)?;
```

That conflates two very different states. A config that is **present on disk but unreadable** (e.g. permission-denied) was reported as "config not found". Because `validate_agent` intentionally *allows* `Missing` (the file may legitimately be created later), an unreadable config slipped silently past the edit-time guard — and then at fire time the routine was just logged-and-skipped. The net effect: a routine shows a green health dot but never actually fires, with no actionable signal about the broken file. (Issue #445.)

## What

- Only `std::io::ErrorKind::NotFound` now maps to `AgentLoadError::Missing`. Any other I/O error yields a new `AgentLoadError::Read(String)` variant carrying the underlying error string.
- `validate_agent` surfaces `Read` as a `400 BadRequest` at edit time, mirroring the existing `Parse` handling, so an unreadable config is rejected up front instead of being waved through.
- The two log-only call sites (`svc_trigger`, `sync/routines.rs`) already render the error via `Display`; the new variant prints `unreadable agent config: <err>`.
- Added a unix test asserting a `chmod 000` config resolves to `Read` (or `Ok` under root, which bypasses the permission bit) rather than `Missing`, plus a `Display` assertion for the new variant.

## Validation

- `cargo build` — clean
- `cargo test agent` (RUST_TEST_THREADS=1) — 31 passed, including the new `load_agent_command_reports_read_error_for_unreadable_config`
- `cargo fmt --check` — clean
- `cargo clippy --all-targets` — no warnings

Scope is one logical fix: distinguishing missing vs. unreadable agent configs. No behavior change for the missing or malformed paths.

@tupe12334

This pull request was opened by the Daemon + Landing auto-improve PRs routine of moadim.


tupe12334:
Blocked: superseded by upstream. `AgentLoadError` in `src/routines/agents/mod.rs` (main) already distinguishes `Missing` from `Unreadable(String)` (I/O error) from `Parse(String)` (TOML error) — exactly this PR's fix. Needs a manual look before merge.

— PR comment commander

tupe12334:
Blocked: superseded. Issue #445 is already fixed on `main` via commit 776c04e (merged PR #527, "fix(agents): distinguish an unreadable agent config from a missing one (#445)") — `AgentLoadError::Unreadable` already exists in `src/routines/agents/mod.rs`. Nothing left to merge — recommend closing as superseded.

— PR comment commander

tupe12334:
Blocked: this branch conflicts with `main`, and looks superseded — `main`'s `src/routines/agents/mod.rs` already has `AgentLoadError` with `Missing`/`Unreadable`/`Parse` variants distinguishing exactly this case. Merging as-is would just reintroduce a duplicate mechanism. Recommend closing as superseded unless there's a gap this PR covers that main's version doesn't.

PR #448, opened 2026-06-20
fix(cron): resolve extensioned handlers on manual trigger (#440)

### LINKED ISSUE #440: Manual trigger (POST /cron-jobs/{id}/trigger) can't find extensioned handlers — diverges from scheduled run's resolve_handler_path, so handlers with a .sh/.py extension silently no-op
## Problem

Manual triggers resolve the handler differently from scheduled runs, so any handler stored **with a file extension** (the documented layout) silently does nothing when triggered.

- Scheduled runs build the crontab line via `sync::resolve_handler_path` (`src/sync/mod.rs:110`), which tries the exact name and then common script extensions: `sh, py, js, rb, pl, bash, zsh`.
- Manual trigger `svc_trigger` (`src/cron_jobs.rs:336`) instead does a bare:
  ```rust
  let handler_path = crate::paths::handlers_dir().join(&job.handler);
  if handler_path.exists() { /* spawn */ } else { log::warn!("trigger: handler script not found ..."); }
  ```
  This only matches an **extensionless** file.

The README documents handlers stored *with* extensions and a `handler` field *without* one:

```
handlers/send-report.sh   ← handler = "send-report"
handlers/cleanup-temp.py  ← handler = "cleanup-temp"
```

For that documented layout:

- Scheduled crontab run → `resolve_handler_path("send-report")` → finds `send-report.sh` → **fires**. ✅
- `POST /cron-jobs/{id}/trigger` (UI "Run now" / REST / MCP) → `handlers_dir().join("send-report")` → `.exists()` is `false` → logs `trigger: handler script not found` and **returns `200` with `last_manual_trigger_at` stamped** → silent no-op. ❌

So the same job that fires correctly on schedule does nothing on manual trigger, and the `200` + updated timestamp make the trigger look successful.

## Motivation

- **Correctness / consistency:** two resolution paths for one concept — scheduled runs honor extensions, manual triggers don't.
- **Silent failure:** the trigger returns success and stamps `last_manual_trigger_at` while the handler never executes; only a `warn`-level log hints at it. This matches the recurring "green dot / 200 but silently no-ops" failure mode already tracked across the repo (e.g. #383, #301, #322).

## Acceptance criteria

- `svc_trigger` resolves the handler through the shared `resolve_handler_path` (or an equivalent shared helper) so extensioned handlers (`.sh`, `.py`, …) are found — identical to scheduled-sync resolution.
- When no handler file resolves, the trigger surfaces a real error (e.g. a `4xx` / `AppError`) instead of returning `200` with only a log line.
- Regression test: a job whose handler is stored as `<name>.sh` with `handler = "<name>"` actually executes on `POST /cron-jobs/{id}/trigger`.
- Optional but recommended: factor handler-path resolution into a single shared helper so the scheduled and manual paths can't drift apart again.

---

_This issue was opened as a suggestion by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. It is intentionally left unassigned — self-assign it to approve. cc @tupe12334_

tupe12334:
## Why

Manually triggering a cron job (`POST /api/v1/cron-jobs/{id}/trigger`) resolved
its handler with a bare `handlers_dir().join(&job.handler)`, which only matches an
**extensionless** filename. The scheduled crontab launch (`format_crontab_line` →
`resolve_handler_path`) instead tries an exact match first, then common script
extensions (`sh`, `py`, `js`, `rb`, `pl`, `bash`, `zsh`).

The two paths disagreed: a handler stored as e.g. `greet.sh` fired correctly on
schedule but the manual trigger looked for exactly `greet`, found nothing, logged
`trigger: handler script not found`, and silently no-opped. Fixes #440.

## What

- `svc_trigger` now routes the handler lookup through the shared
  `crate::sync::resolve_handler_path`, so manual and scheduled launches resolve
  identically. The existing `.exists()` / best-effort-spawn behaviour is unchanged.
- Adds a regression test (`svc_trigger_resolves_extensioned_handler`) that stores
  an extensionless handler name backed by a `.sh` file on disk and asserts the
  trigger takes the spawn branch.

## Verification

```
cargo fmt --check
cargo clippy
cargo test svc_trigger
cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'
```

All pass locally (100% line coverage preserved). CHANGELOG updated under
`[Unreleased] → Fixed`.

---


tupe12334:
Blocked: superseded by upstream. `svc_trigger` and `svc_trigger_scheduled` both delegate to the single `spawn_routine_command(&routine)` in `src/routines/service.rs` — there's no separate handler-path-resolution logic left to diverge on manual trigger. Needs a manual look (likely just closing).

— PR comment commander

Issue #450, opened 2026-06-20
cron-jobs sync_to_crontab lacks the empty-store wipe guard that routines have — a failed/empty store load clobbers all managed job lines on the next mutation


tupe12334:
## Problem

`sync_to_crontab` (the **cron-jobs** forward sync, `src/sync/mod.rs`) writes the freshly built managed block over the live crontab with no protection against an empty store:

```rust
pub fn sync_to_crontab(store: &CronStore) -> Result<(), SyncError> {
    let current = read_crontab()?;
    let block = build_block(store);              // empty store -> bare block
    let new_crontab = replace_block(&current, &block);
    if new_crontab == current { return Ok(()); }
    write_crontab(&new_crontab)                  // wipes every managed job line
}
```

The **routines** sibling, `sync_routines_to_crontab`, already guards exactly this case (added after a real incident — see its `// Footgun guard` comment):

```rust
if store.lock().unwrap().is_empty() && current.contains(ROUTINE_LINE_MARKER) {
    log::warn!("...store is empty but the crontab still has routine lines; refusing to wipe...");
    return Ok(());
}
```

The cron-jobs path never received this protection. The danger is concrete because the store can come up **empty for a non-genuine reason**: `load_store_from_dir` (`src/storage.rs`) swallows a `read_dir` failure and returns an empty store —

```rust
if let Ok(entries) = std::fs::read_dir(dir) { ... }   // Err -> empty store, no log
Arc::new(Mutex::new(jobs))
```

So a transient/permission error reading `~/.config/moadim/jobs/`, or a second daemon racing this one, yields an empty `CronStore`. The next `svc_create` / `svc_update` / `svc_delete` (all three call `sync_to_crontab`) then builds a bare block and silently strips **every** managed cron-job line from the user's live crontab — the jobs stop firing until the next successful mutation re-populates the block.

## Motivation

- Data-loss / silent-failure parity gap: routines are protected against the empty-store wipe; cron jobs are not, despite sharing the same forward-sync shape and the same load path that can fail to empty.
- Distinct from existing issues: #324 is the marker-collision wipe of the *routines* block, #365 is the missing read-modify-write lock (lost update between the two blocks), and #394 is startup *not re-syncing* a lost block. None of them guard `sync_to_crontab` against an empty/failed-load store clobbering the jobs block.

## Nuance (don't break "delete my last job")

[...]

PR #460, opened 2026-06-20
fix(sync): guard cron-jobs sync against wiping the block on an empty store (#450)

tupe12334:
## Why

`sync_to_crontab` (the cron-jobs forward sync) rebuilt the managed-jobs crontab block from the store on **every** mutation, with no guard. If the store was empty at sync time — a failed or partial load on startup, or a second daemon racing this one — `build_block` emitted a bare block and the sync overwrote the crontab, **silently dropping every managed job line** until the next successful mutation re-populated it.

The routines block already defends against this exact footgun in `sync_routines_to_crontab` (added for the routines incident). The cron-jobs path was missing the equivalent guard — issue #450.

## What

Mirror the routines guard in `sync_to_crontab`:

- When the store is empty **and** the crontab still carries managed job lines (`# moadim:`), log a warning and skip the write.
- A store that loaded fine but legitimately holds only disabled/unmanaged jobs is *not* empty, so clearing the last job still works as before.

The new `JOB_LINE_MARKER` (`# moadim:`) does not match the routines marker (`# moadim-routine:`), so the two blocks stay independent.

## Tests

Adds two tests covering both guard branches, matching the routines-guard test pair:

- `sync_refuses_to_wipe_job_lines_when_store_is_empty` — populated block + empty store ⇒ crontab untouched.
- `sync_proceeds_when_store_empty_but_no_job_lines` — empty store, no job marker ⇒ guard's second operand is false, sync proceeds (idempotent no-op).

## Verification (local)

- `cargo clippy --all-targets --all-features -- -D warnings` — clean.
- `cargo test sync::cron_sync_tests::sync_refuses_to_wipe_job_lines_when_store_is_empty sync::cron_sync_tests::sync_proceeds_when_store_empty_but_no_job_lines` — `2 passed; 0 failed`.

---

This pull request was opened by the *Daemon + Landing auto-improve PRs* routine of moadim. @tupe12334

### DIFF
--- a/src/sync/cron_sync_tests.rs
+++ b/src/sync/cron_sync_tests.rs
@@ -540,6 +540,35 @@ fn sync_to_crontab_writes_block_and_is_idempotent() {
     drop(shim);
 }
 
+#[test]
+fn sync_refuses_to_wipe_job_lines_when_store_is_empty() {
+    // Footgun guard (#450): an empty store must NOT overwrite a populated job block — that would
+    // silently drop every managed cron line. Mirrors the routines-block guard.
+    let populated = "\
+# BEGIN MOADIM
+# Managed by moadim
+30 9 * * 1-5 /handlers/keep-me # moadim:keep-me
+# END MOADIM
+";
+    let shim = CronShim::new(Some(populated));
+    sync_to_crontab(&store_with(vec![])).unwrap();
+    // The crontab is left untouched: the managed job line survives.
+    assert_eq!(shim.store_contents(), populated);
+    assert!(shim.store_contents().contains("# moadim:keep-me"));
+    drop(shim);
+}
+
+#[test]
+fn sync_proceeds_when_store_empty_but_no_job_lines() {
+    // Guard's second operand is false (no `# moadim:` marker), so sync proceeds; the block is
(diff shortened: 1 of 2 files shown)


tupe12334:
Blocked: superseded by upstream. The empty-store wipe guard is already implemented in `sync_routines_to_crontab` (`src/sync/routines.rs`), which checks for an empty store + existing marker before writing. Needs a manual look (likely just closing).

— PR comment commander

tupe12334:
Blocked: superseded. Issue #450 is already closed by merged PR #542 ("fix(cron): guard cron-jobs sync against wiping the block on an empty store (#450)"), already on `main`. Nothing left to merge — recommend closing as superseded.

— PR comment commander

tupe12334:
Blocked: this branch conflicts with `main` because the feature it targets is gone — `main` removed the whole cron-jobs REST feature in `c7ad088` "feat!: remove cron jobs, focus on AI-agent routines (#842)", including `src/sync/cron_sync_tests.rs` and the cron-jobs sync path this PR guards. There's nothing left to apply this fix to. Recommend closing as obsolete.

PR #484, opened 2026-06-20
fix(stores): recover poisoned store locks instead of bricking the daemon (#363)

### LINKED ISSUE #363: Store Mutex poisoning permanently bricks the daemon: one panic under any held lock makes every routine/cron endpoint panic until restart
## Problem

The shared in-memory stores are guarded by `std::sync::Mutex` and every one of the ~25 access sites takes the lock with `.lock().unwrap()`:

```
pub type CronStore = Arc<Mutex<HashMap<String, CronJob>>>;
// e.g. src/cron_jobs.rs, src/routines/service.rs, src/sync/*.rs, src/routines/cleanup/*.rs, ...
let lock = store.lock().unwrap();
```

A `std::sync::Mutex` becomes **poisoned** the moment any thread panics while holding the guard. After that, *every* subsequent `.lock().unwrap()` returns `Err(PoisonError)` and the `.unwrap()` panics. Because the `CronStore` / `RoutineStore` are process-wide singletons, a single panic under any held lock permanently bricks the whole daemon: list/get/create/update/delete/trigger for both cron-jobs and routines, the iCal feed, crontab sync, and the periodic workbench cleanup sweep all panic on their next lock acquisition. The only recovery is a full process restart.

Confirmed there is **no** poison handling anywhere in the codebase: no `clear_poison`, no `into_inner`, no `unwrap_or_else(|e| e.into_inner())`, and no `catch_unwind` / panic-recovery layer.

This is distinct from #337 (add an HTTP panic-recovery layer). A per-request `catch_unwind` layer would convert the *first* panic into a 500, but the lock stays poisoned, so every later request that touches the store still panics — #337 does not un-poison the mutex. The two issues are complementary; this one is about the store's permanent-failure mode, not the response code of the request that triggered it.

## Motivation

- A long-lived, network-exposed daemon should degrade gracefully, not enter a permanent all-endpoints-down state from one stray panic (e.g. a future `unwrap` on malformed data while a lock is held).
- The failure is silent and confusing: after poisoning, healthy-looking requests start returning connection resets / 500s with no obvious cause, and `/health` (which doesn't lock the store) still reports `ok`.
- The fix is small, localized, and removes a foot-gun that grows with every new `.lock().unwrap()` added.

## Acceptance criteria

- [ ] Store lock acquisition no longer panics on a poisoned mutex. Pick one approach and apply it consistently:
  - recover the guard via `unwrap_or_else(|e| e.into_inner())` behind a small helper (e.g. `fn lock_store(&self) -> MutexGuard<..>`), **or**
  - swap the guard type for a non-poisoning mutex (e.g. `parking_lot::Mutex`), **or**
  - document and justify keeping poisoning if intentional.
- [ ] All ~25 `.lock().unwrap()` call sites across `cron_jobs.rs`, `routine_storage.rs`, `routines/service.rs`, `routines/defaults.rs`, `routines/ical.rs`, `routines/cleanup/*`, and `sync/*` use the chosen non-panicking path.
- [ ] A regression test poisons a store lock (spawn a thread that panics while holding it, join it) and asserts that a subsequent service-layer call still succeeds instead of panicking.
- [ ] No new `.lock().unwrap()` is reintroduced (consider a `clippy.toml` disallowed-method entry or a grep check in CI).

---
*This issue was suggested by the "Suggest new GitHub issues (daemon + landing)" routine of moadim — filed as an unassigned suggestion. Approve it by self-assigning. cc @tupe12334*

tupe12334:
## Problem

The shared `CronStore` / `RoutineStore` are process-wide `Arc<Mutex<HashMap>>` singletons accessed with `.lock().unwrap()` at ~20 production sites. A `std::sync::Mutex` poisons **permanently** the instant any thread panics while holding the guard, after which every later `.lock().unwrap()` panics too. Because the stores are singletons, one stray panic under any held lock bricks list/get/create/update/delete/trigger for both cron-jobs and routines, the iCal feed, crontab sync and the cleanup sweep — recoverable only by a full restart, while `/health` (which never locks the store) keeps reporting `ok`.

## Fix

- Add a `LockRecover` extension trait in `src/utils/lock.rs` whose `lock_recover()` acquires the guard via `self.lock().unwrap_or_else(|e| e.into_inner())`.
- Route all ~20 production store accesses (`cron_jobs.rs`, `routine_storage.rs`, `routines/service.rs`, `routines/ical.rs`, `routines/defaults.rs`, `routines/cleanup/snapshot.rs`, `sync/routines.rs`, `sync/mod.rs`) through `.lock_recover()`.

The data behind the lock is still a structurally valid `HashMap`, so after a poisoning panic the daemon degrades to an at-worst partially-applied mutation rather than a dead process. The named helper also makes any future bare `.lock().unwrap()` stand out in review.

## Tests

Added `src/utils/lock_tests.rs`: a thread panics while holding the guard (poisoning the mutex), and a subsequent `lock_recover()` is asserted to return the inner data instead of panicking; plus a healthy-path test.

- `cargo fmt --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `cargo test` ✅ (488 passed)

Closes #363

---
🤖 This pull request was opened by the **Open issues → fix PRs (owned repos + my orgs)** routine of moadim. cc @tupe12334


tupe12334:
Blocked: superseded by upstream. `LockRecover`/`lock_recover()` (`src/utils/lock.rs`) already exists and is used pervasively across the store-locking call sites this PR would have touched. Needs a manual look (likely just closing).

— PR comment commander

tupe12334:
Blocked: superseded. Issue #363's poison-tolerant locking is already shipped via merged PR #496 ("refactor: remove panicking unwrap from store access") — `src/utils/lock.rs`'s `LockRecover::lock_recover()` already covers every store access site. Nothing left to merge — recommend closing as superseded.

— PR comment commander

tupe12334:
Blocked: this branch conflicts with `main`, and looks superseded — `main` already has the same fix via `src/utils/lock.rs`'s `LockRecover` trait (`lock_recover()`, recovering the guard instead of panicking on a poisoned mutex), applied at every store call site (no `.lock().unwrap()`/`.expect()` remain in `src/`). Merging as-is would just reintroduce a duplicate mechanism. Recommend closing as superseded unless there's a call site this PR covers that main's version missed.

PR #434, opened 2026-06-19
fix(ui): box the large Routine variant in RAction to satisfy clippy

tupe12334:
## Why

The `ui` crate enforces `[lints.clippy] all = "deny"`, yet `cargo clippy --all-targets` currently **fails to compile** it on current stable (cargo 1.96.0):

```
error: large size difference between variants
   --> ui/src/routines.rs:281:1
    |
293 |     Upsert(Routine),
    |     --------------- the largest variant contains at least 272 bytes
    | the second-largest variant contains at least 48 bytes
    = note: `-D clippy::large-enum-variant` implied by `-D clippy::all`
```

`RAction::Upsert` held a whole `Routine` (~272 bytes) by value while every other variant is ≤48 bytes, so each dispatched action paid for the largest variant on the stack. This is a real build break for any UI work and for the lint gate the crate opts into.

## What

- `Upsert(Routine)` → `Upsert(Box<Routine>)`.
- Reducer derefs once (`let routine = *routine;`) before the unchanged upsert logic.
- Wrap the four `RAction::Upsert(r)` construction sites in `Box::new(r)`.

## Verification

```
cargo clippy --all-targets   # clean
cargo fmt --check            # clean
```

Focused, behavior-preserving change — no API or UI-behavior difference.

---
This pull request was opened by the **Daemon + Landing auto-improve PRs** routine of moadim. @tupe12334

Issue #155, opened 2026-06-17
Add last_scheduled_trigger_at to track cron-driven runs


tupe12334:
Related: #128 (rename `last_triggered_at` → `last_manual_trigger_at` tracked separately there).

## Problem

`Routine` (and `CronJob`) expose `last_triggered_at`, but it only records **manual** triggers:

> Unix timestamp (seconds) when the routine was last manually triggered, if ever.

No field records when a routine last fired via its **cron schedule**. You cannot tell from the API/model whether (or when) a scheduled run executed — only manual triggers leave a timestamp.

## Proposal

Add `last_scheduled_trigger_at: Option<u64>`, set by the scheduler each time the cron expression fires a run. Distinct from `last_triggered_at` (manual).

## Scope

- `src/routines/model.rs` — add field to `Routine`
- `src/cron_jobs*` — same for `CronJob`; set timestamp on each scheduled fire
- `apis/openapi.json` — document new property (mirror `last_triggered_at`)
- `ui/src/routines.rs`, `ui/src/cron_jobs.rs` — surface field
- Persistence (routine_store) + tests

## Benefit

Distinguish scheduled vs manual execution; detect schedules that never fired; show "last run" in UI.

tupe12334:
Looked at implementing this and hit an architecture mismatch worth confirming before building it.

The proposal says `last_scheduled_trigger_at` would be "set by the scheduler each time the cron expression fires." But there is **no in-process scheduler** — scheduled runs execute via the host **OS crontab**, which invokes `run.sh` (`build_routine_command`) directly; the daemon is not in the loop at fire time. So nothing daemon-side observes a scheduled firing to stamp a timestamp.

Two consequences:

1. **Routines (managed):** the only faithful way to record a scheduled fire is to have the generated `run.sh` itself write the epoch into a dedicated gitignored sidecar (e.g. `scheduled.local.toml`, matching `*.local.*`), which the daemon then reads on load — alongside the existing `state.local.toml` that holds `last_manual_trigger_at`. Workable, but it's a new fire-time-write mechanism, not just a model field.
2. **Cron jobs:** these can be `System` (read-only entries *discovered* from the host crontab — the daemon does not generate their command), so there is **no place to inject a timestamp-write** for those. The scheduled timestamp would only ever be populatable for `Managed` jobs.

I'd rather confirm the approach than force a design. Proposed scope if you're happy with it:
[...]

PR #506, opened 2026-06-21
feat(ui): surface last scheduled fire in the routines table

### LINKED ISSUE #487: UI: surface last_scheduled_trigger_at (last/never fired) in the routines table — #155's data is API-only
## Problem

#155 (completed) added **`last_scheduled_trigger_at`** to the routine REST/OpenAPI response — the scheduled-fire mirror of `last_manual_trigger_at` — with the explicit goal of letting an operator *"spot schedules that have never actually fired."* That data now exists in the API but **never reaches the daemon's own web UI**:

- `ui/src/routines.rs` deserializes only `last_manual_trigger_at` (the `Routine` struct has no `last_scheduled_trigger_at` field — the JSON key is silently dropped).
- The per-row `last_run` value is derived **manual-only**:
  ```rust
  let last_run = r
      .last_manual_trigger_at
      .map(|t| format!("↻ {}", reltime(t)))
      .unwrap_or_default();
  ```
- The routines table columns are `TITLE · SCHEDULE · AGENT · REPOS · TTL · ENABLED · UPDATED` — there is no "last fired" column at all.

**Consequence:** a routine that fires perfectly on its cron schedule but was never manually triggered shows a *blank* last-run cell — visually indistinguishable from a routine that has never run. The operator cannot tell a healthy schedule from a dead one in the UI, which is precisely the diagnosis #155 was built to enable. The whole loop-engine value prop is "set the loop, forget the keyboard" — but the keyboard-free (scheduled) runs are the ones the dashboard hides.

## Motivation

- Surfaces the most operationally important signal: *is this schedule actually firing?* A silently-dropped routine (uninstalled agent, bad PATH, clamped TTL, etc. — see the many silent-no-op issues) currently looks identical to a working one in the UI.
- Completes #155 end-to-end: the field was added to the API on purpose; the dashboard is the natural consumer and was left behind.
- Cheap, well-scoped, and self-contained to `ui/src/routines.rs` (plus the existing `reltime` helper).

## Acceptance criteria

- [ ] The UI `Routine` struct deserializes `last_scheduled_trigger_at: Option<u64>` (with `#[serde(default)]`, matching the other optional fields).
- [ ] The routines table surfaces the last scheduled fire — either a dedicated "LAST FIRED" column or by folding scheduled + manual into a single "last run" cell that shows whichever is most recent, with the source distinguished (e.g. `⏱ 2h ago` scheduled vs `↻ 5m ago` manual).
- [ ] A routine that has **never fired on schedule and was never manually triggered** is visibly flagged as never-run (e.g. a muted "never fired" badge), not a blank cell — so dead schedules stand out.
- [ ] Behavior degrades gracefully when the field is absent (e.g. the bare `Routine` returned by `/trigger`), exactly like the existing derived fields.
- [ ] No backend change required — the data is already in the `/api/v1/routines` response.

## Notes

This is a **new follow-up** to the completed #155, not a duplicate: #155 was API-only. Distinct from #438 (live `is_running` status), #369 (`next_run_at`, the *next* fire), and #477 (retention countdown in the LOGS view).

---
*This issue was opened as a suggestion by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. It is intentionally left unassigned — approve it by self-assigning.*

tupe12334:
## What

The routines table derived its per-row "last run" cell from `last_manual_trigger_at` **only**, so a routine that fires perfectly on its cron schedule but was never triggered manually showed a *blank* cell — visually indistinguishable from one that has never run. That hid exactly the signal #155 added `last_scheduled_trigger_at` to the API to expose: *is this schedule actually firing?*

## Changes

- The UI `Routine` struct now deserializes `last_scheduled_trigger_at: Option<u64>` with `#[serde(default)]`, matching the other optional fields (graceful when absent, e.g. the bare `Routine` returned by `/trigger`).
- New pure `last_fired(scheduled, manual)` helper folds the two timestamps into the most-recent fire, tagged with a source icon — `⏱` scheduled vs `↻` manual. Ties favor the scheduled source.
- A routine that has **never fired on schedule and was never manually triggered** now renders a muted italic **"never fired"** badge (new `.cell-triggered.never` style) instead of a blank cell, so dead schedules stand out.
- No backend change — the data is already in the `/api/v1/routines` response. Completes #155 end-to-end (which was API-only).

## Notes

`cargo clippy -p ui` currently fails on an unrelated pre-existing `large_enum_variant` lint on `RAction::Upsert(Routine)` (fails on clean `main` too, on current stable) — already tracked by open PR #434 and out of scope here. The code added by this PR is clippy-clean; `cargo fmt --check` passes.

This pull request was opened by the "Implement my assigned issues without a PR (daemon + landing)" routine of moadim.

Closes #487


tupe12334:
Blocked: superseded by upstream. `ui/src/routines/row.rs` (main) already folds `last_manual_trigger_at`/`last_scheduled_trigger_at` into a single cell with a source icon (`↻` manual / `⏱` scheduled), matching this PR's core ask. Needs a manual look (likely just closing).

— PR comment commander

PR #513, opened 2026-06-21
fix(openapi): relative servers URL so Swagger "Try it out" follows the origin

### LINKED ISSUE #385: OpenAPI `servers` URL is hardcoded to 127.0.0.1:5784 — Swagger "Try it out" breaks under MOADIM_BIND_ADDR / reverse proxy
## Problem

The OpenAPI spec hardcodes a single absolute `servers` URL:

```rust
// src/openapi.rs:6
servers((url = "http://127.0.0.1:5784/api/v1", description = "Local development")),
```

This is baked into both the live spec served at `/docs/openapi.json` and the committed `apis/openapi.json`. The Swagger UI mounted at `/docs` uses this `servers` entry as the base URL for every **"Try it out"** request.

The bind address is **not** fixed: `MOADIM_BIND_ADDR` (`src/cli.rs`) lets an operator move the daemon to a different host or port. When it does, the Swagger UI keeps pointing "Try it out" at `http://127.0.0.1:5784/api/v1`, so requests issued from `/docs` hit the wrong port (connection refused) or, worse, a *different* service listening on `5784`. The interactive docs are silently broken for any non-default bind.

It also breaks the common reverse-proxy / remote case: the page loads fine over the proxied origin, but every "Try it out" call fires at loopback `5784` instead of the origin the browser actually reached the UI on.

## Motivation

- `/docs` "Try it out" is the main reason to ship an embedded Swagger UI; pointing it at a hardcoded address defeats the feature whenever `MOADIM_BIND_ADDR` is set.
- This mirrors the already-fixed class of "hardcoded default address" bugs (#171) and the OpenAPI-drift bugs (#309 version, #319 spec path) — same root cause (compile-time constant where a runtime/origin-relative value belongs), different surface.

## Proposed fix

Use a **relative** server URL so Swagger resolves "Try it out" against the page's own origin:

```rust
servers((url = "/api/v1", description = "This server")),
```

A relative `url` is valid per OpenAPI 3.0 (resolved against the document's retrieval URL), so it works on loopback, on a custom `MOADIM_BIND_ADDR`, and behind a reverse proxy with no rebuild. Regenerate `apis/openapi.json` to match.

## Acceptance criteria

- [ ] OpenAPI `servers[0].url` is `/api/v1` (relative), not an absolute `http://127.0.0.1:5784/...`.
- [ ] Loading `/docs` on a non-default `MOADIM_BIND_ADDR` (e.g. `127.0.0.1:6000`) and using "Try it out" sends requests to the actual bound address, not `:5784`.
- [ ] `apis/openapi.json` is regenerated and committed so the on-disk spec matches.
- [ ] A test asserts the generated spec's `servers[0].url` is relative (guards against re-introducing a hardcoded absolute URL).
- [ ] Out of scope: the `info.version` drift (tracked in #309) and the spec-rewrite path behavior (#319).

---
_This issue was suggested by the "Suggest new GitHub issues (daemon + landing)" routine of moadim. It is a suggestion only and left intentionally unassigned — self-assign to approve._

tupe12334:
## Problem

The OpenAPI spec hardcoded a single absolute `servers` URL:

```rust
servers((url = "http://127.0.0.1:5784/api/v1", description = "Local development")),
```

Swagger UI at `/docs` uses that entry as the base URL for every **"Try it out"** request. The bind address isn't fixed (`MOADIM_BIND_ADDR`), so on any non-default bind — or behind a reverse proxy — the interactive docs kept firing requests at loopback `:5784` (connection refused, or a *different* service on that port). The docs are silently broken off the default address.

## Fix

Use a **relative** server URL, valid per OpenAPI 3.0 (resolved against the document's own retrieval origin):

```rust
servers((url = "/api/v1", description = "This server")),
```

Now "Try it out" follows whatever address served the page — loopback, a custom `MOADIM_BIND_ADDR`, or a proxied origin — with no rebuild.

- Regenerated and committed `apis/openapi.json` (only the `servers[0]` entry changes).
- Added `server_url_is_relative` test guarding against re-introducing a hardcoded absolute URL.
- Added a CHANGELOG `[Unreleased]` entry.

Out of scope (tracked separately): `info.version` drift (#309), spec-rewrite path behavior (#319).

## Verification

`cargo test openapi`, `cargo fmt --check`, and `cargo clippy --all-targets -- -D warnings` all pass.

Closes #385

---
This pull request was opened by the "Open issues → fix PRs (owned repos + my orgs)" routine of moadim.


tupe12334:
Blocked: superseded by upstream. `src/openapi.rs` (main) already declares the servers URL as the relative `/api/v1`, not an absolute origin. Needs a manual look before merge.

— PR comment commander

PR #519, opened 2026-06-21
fix(ui): bundle the Share Tech Mono webfont inline instead of fetching it from Google's CDN (#467)

### LINKED ISSUE #467: Served UI fetches its webfont from Google's CDN at runtime — the self-contained daemon UI breaks offline, leaks each load to a third party, and flashes on first paint
## Problem

The served web UI loads its monospace webfont (`Share Tech Mono`) from **Google's CDN at runtime**. The actually-shipped, embedded `prebuilt.html` (not just the dev `ui/index.html`) contains:

```html
<link rel="preconnect" href="https://fonts.googleapis.com" />
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="" />
<link href="https://fonts.googleapis.com/css2?family=Share+Tech+Mono&display=swap" rel="stylesheet" />
```

and the whole UI is themed `font-family: 'Share Tech Mono', 'Courier New', monospace;`.

moadim ships as a **self-contained single binary** — `prebuilt.html` is embedded and served by the daemon (the install story is `cargo install moadim && moadim`, and #13 deliberately bundled the UI into the binary so `cargo install` ships the UI). Pulling the font from `fonts.googleapis.com` / `fonts.gstatic.com` on every UI load breaks that self-contained promise:

- **Offline / air-gapped hosts** (the daemon binds `127.0.0.1` by default and is exactly the kind of thing run on headless or network-restricted boxes) silently fall back to `Courier New`/`monospace` — the intended CRT aesthetic is lost.
- **Privacy / third-party leak**: every operator opening the localhost UI makes a request to Google's CDN, leaking visit timing/IP/User-Agent. This is at odds with the project's care about not leaking internals (e.g. #356 stripping `x-server-root` path headers).
- **Render flash / latency (FOUT)**: `display=swap` plus a remote fetch means the UI repaints once the font arrives, and the first paint stalls on a third-party round-trip.

## Motivation

A localhost control panel for a self-hosted daemon should render identically with the network unplugged. The font is the only runtime external dependency of the served page; removing it makes the UI fully self-contained, faster, and private — consistent with the single-binary install model.

## Acceptance criteria

- [ ] The served UI renders with its intended font and **zero requests to `fonts.googleapis.com` / `fonts.gstatic.com`** (verify in DevTools Network tab with the page loaded offline).
- [ ] Self-host the `Share Tech Mono` `woff2` instead of the CDN: either embed it in `prebuilt.html` via a base64 `@font-face` `src: url(data:font/woff2;base64,…)`, or bundle it as a Trunk asset served by the daemon, with an `@font-face` rule pointing at the local path.
- [ ] Drop the two `preconnect` links and the Google Fonts `<link rel="stylesheet">` from both `ui/index.html` and the generated `prebuilt.html`.
- [ ] Confirm the font's license (SIL OFL for Share Tech Mono) permits embedding/self-hosting, and include the license/attribution as required.
- [ ] The build pipeline regenerates `prebuilt.html` with the self-hosted font so a fresh `cargo install moadim` ships it (no manual step).
- [ ] The `monospace` fallback stack is retained so a missing/locally-unloadable font still degrades gracefully.

---
_This issue was filed as a non-binding suggestion by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. It is intentionally left unassigned — self-assign it to approve. cc @tupe12334_

tupe12334:
## Why

The served daemon UI is built to be **self-contained** — the whole SPA (HTML, CSS, JS, and the base64-inlined WASM) is collapsed into a single file and shipped inside the binary so `cargo install` users get a working dashboard with no external assets. But the UI `<head>` still pulled its `Share Tech Mono` typeface from `fonts.googleapis.com` at runtime, so the "self-contained" UI:

- **leaked to a third party** — every dashboard load fired a request to Google's font CDN, so merely opening your own daemon UI tells Google about it;
- **broke offline** — on an air-gapped/offline host the intended typeface never loaded; and
- **flashed** — a fallback font showed on first paint until the webfont finished downloading (FOUT).

## What

- Replace the three `<link>` tags in `ui/index.html` (two `preconnect`s + the stylesheet) with an inline `@font-face` whose `src` is the woff2 embedded as a base64 `data:` URI (~13.5 KB). The font now ships with the page: **no network request, correct typeface offline, no first-paint swap.**
- The font is **Share Tech Mono**, SIL Open Font License 1.1 — attributed in a CSS comment next to the `@font-face`. OFL permits embedding.
- Patch `prebuilt.html` (the committed, shipped artifact that `cargo install` serves, regenerated from `ui/index.html` by the build script via trunk) **in lockstep** — only the font lines change; the compiled WASM/JS bytes are untouched, keeping the diff reviewable instead of a full-rebuild churn.

## Notes

- The daemon's CSP (`src/middlewares/security_headers.rs`) only sets `frame-ancestors 'none'` — it does not constrain `font-src`/`style-src`, so the inline `data:` font loads with **no header change** (the previous remote font relied on the same absence).
- Verified the embedded base64 decodes byte-for-byte to the upstream woff2, and that no `https://fonts.*` reference remains in either file.
- No Rust changed; `cargo fmt --check` passes.

Closes #467

---
This pull request was opened by the **Daemon + Landing auto-improve PRs** routine of moadim. /cc @tupe12334


tupe12334:
Blocked: superseded by upstream. `ui/index.html` (main) already self-hosts Share Tech Mono via a local `@font-face`, with zero runtime requests to fonts.googleapis.com/fonts.gstatic.com. Needs a manual look before merge.

— PR comment commander

PR #527, opened 2026-06-21
fix(agents): distinguish an unreadable agent config from a missing one (#445)

tupe12334:
## Why

`load_agent_command` mapped **every** `read_to_string` failure to `AgentLoadError::Missing`:

```rust
std::fs::read_to_string(agent_toml_path(name)).map_err(|_| AgentLoadError::Missing)?;
```

So an agent config that *exists on disk but cannot be read* (a permissions error, or a path that is a directory) was reported as "agent config not found". That label is misleading on its own, but it also changes behavior: `validate_agent` intentionally accepts `Missing` (the file may be created later), so an unreadable config **passed validation** and left a green-dot routine that silently never fires — exactly the failure mode described in #445.

## What

- Add `AgentLoadError::Unreadable(String)` and only map a genuine `ErrorKind::NotFound` to `Missing`; any other I/O error now carries its underlying cause.
- `validate_agent` rejects `Unreadable` with a `BadRequest` (mirroring the existing malformed-config arm), so the operator learns the real cause at create/update time instead of getting a routine that never runs.
- `svc_trigger` and the crontab-sync warning lines improve automatically — they already render the error via `Display`, which now says "unreadable agent config: …" instead of "agent config not found".

## Tests

- `load_agent_command_reports_unreadable_for_non_not_found_io_error` — exercises the new load branch portably (a directory at the config path yields a non-`NotFound` I/O error on every platform; no permission-bit tricks that root would bypass).
- `svc_create_rejects_unreadable_agent_config` — covers the new validation rejection.

Existing `Missing` / `Parse` tests are unchanged.

## Checks

- `cargo fmt --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `cargo test` (agents + service suites) ✅

---


tupe12334:
LGTM

tupe12334:

tupe12334:
Heads up: this PR is approved (LGTM) and content-mergeable, but the merge is **BLOCKED** by a failing required check: `unreleased-entry` (fail). I will not force-merge past a failing required check. Once the changelog/unreleased entry is added and the check passes, it can be merged.

PR #529, opened 2026-06-21
fix(cron): resolve extensioned handlers on manual trigger (#440)

### LINKED ISSUE #440: Manual trigger (POST /cron-jobs/{id}/trigger) can't find extensioned handlers — diverges from scheduled run's resolve_handler_path, so handlers with a .sh/.py extension silently no-op

tupe12334:
## Why

`POST /cron-jobs/{id}/trigger` (`svc_trigger`) resolved the handler differently from the scheduled run, so a manual trigger could silently no-op on a job that fires fine on schedule.

- **Scheduled run** builds the crontab line via `resolve_handler_path`, which tries an exact match and then common script extensions (`.sh`, `.py`, `.js`, `.rb`, `.pl`, `.bash`, `.zsh`).
- **Manual trigger** did `handlers_dir().join(&job.handler)` — a bare join with no extension fallback.

So a job stored as `send-report` but backed by `send-report.sh` on disk:
- fires correctly on schedule ✅
- on manual trigger logs `trigger: handler script not found at …/send-report` and does nothing ❌

This is the divergence described in #440.

## What

- `svc_trigger` now resolves the handler through the same `crate::sync::resolve_handler_path` the scheduled path uses, so both agree.
- Added a regression test (`svc_trigger_resolves_extensioned_handler`): creates a job whose `handler` is stored **without** an extension but backed by `<name>.sh`, triggers it, and asserts the script actually ran (it `touch`es a sentinel file). The test fails against the old bare-join behavior.

## Checks

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (517 passed) all green locally on stable 1.96.

Closes #440.

---


tupe12334:
Blocked: superseded by upstream. Duplicate concern of #448 — manual-trigger handler resolution is now a single shared code path, nothing left to diverge. Needs a manual look (likely just closing).

— PR comment commander

PR #538, opened 2026-06-21
test(cli): pin the status/stop --json key-parity contract

tupe12334:
## Why

`status --json` and `stop --json` are documented (in their doc comments) to share **one** object shape — `{"running", "pid", "address"}` — so a script can parse either uniformly. Today that's only enforced by independent per-field tests on `status_json` and `stop_json`. Adding a field to one and forgetting the other would pass every existing test and silently break the parity that consumers rely on.

## What

Adds a single `cli_tests` assertion (`status_and_stop_json_share_the_same_keys`) that compares the two functions' top-level key **sets** directly, in both the running and not-running states. If the shapes ever drift, this test fails at the source instead of surfacing in a downstream `--json` consumer.

This implements an item from `TODO.md` ("Add a `cli_tests` assertion that `status --json` and `stop --json` produce the SAME set of object keys") — the line is removed in this PR per the file's stated convention.

## Notes
- Test-only; no production code changes.
- Local gates pass: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test`.

---


tupe12334:
Blocked: superseded by upstream. `src/cli_tests.rs` (main) already pins the status/stop `--json` key-parity contract via `json_key_set` + `is_subset` assertions. Needs a manual look (likely just closing).

— PR comment commander

Issue #128, opened 2026-06-17
Misleading field name: `last_triggered_at` only records manual triggers, not scheduled runs


tupe12334:
## Problem

`last_triggered_at` is only updated on **manual** triggers. `svc_trigger` (`src/routines/service.rs:154-157`) sets `routine.last_triggered_at = Some(now_secs())`, but that path runs only for the manual `trigger_routine` MCP/HTTP action.

Scheduled (cron) firings execute the built command directly via crontab and never go through `svc_trigger`, so they do **not** update the field. The name `last_triggered_at` implies *any* trigger — scheduled included — which is wrong and misleading.

## Impact

- A routine that fires on schedule every 5 min but was never manually triggered shows `last_triggered_at: null`, reading as "never ran" when it has run many times.
- Anyone using the field to answer "when did this last run?" gets the wrong answer.

## Options

1. **Rename** to reflect reality: `last_manual_trigger_at` (and update MCP tool descriptions in `src/routes/mcp.rs` that say "recording last_triggered_at").
2. **Fix the semantics**: also update the timestamp on scheduled runs so the name becomes accurate (requires the scheduled command path to write it back — bigger change).
3. Keep `last_triggered_at` for manual + add a separate `last_run_at` for scheduled.

## Affected
- `src/routines/service.rs:154-157`
- `src/routines/model.rs:48`, `src/routine_storage.rs`
- `src/cron_jobs.rs:323-328` / `src/storage.rs` — same pattern for cron jobs
- MCP tool descriptions: `src/routes/mcp.rs:240,321`

Related: #127

tupe12334:
I want to rename for last_manual_trigger_at and also have last_scheduled_trigger_at

PR #541, opened 2026-06-21
fix(routines): don't stamp the scheduled-fire sidecar on a manual trigger (#478)

### LINKED ISSUE #478: Manual trigger pollutes last_scheduled_trigger_at: svc_trigger runs the shared command, which writes the scheduled-fire sidecar — conflating manual and scheduled runs (#128/#155)
## Problem

A **manual** trigger (`POST /routines/{id}/trigger` → `svc_trigger`) stamps the routine's **scheduled** fire time, falsely recording that the cron schedule fired when it did not.

`svc_trigger` runs the exact same command the crontab runs, by design:

```rust
// src/routines/service.rs (svc_trigger)
let cmd = build_routine_command(&routine, &agent);
std::process::Command::new("sh").arg("-lc").arg(&cmd).spawn()
```

But `build_routine_command` *unconditionally* emits a statement that writes the scheduled-state sidecar:

```rust
// src/routines/command.rs (build_routine_command)
// "Record this scheduled firing."
format!(
    r#"printf 'last_scheduled_trigger_at = %s\n' "$TS" > {} || true"#,
    shell_quote(&scheduled_state_path)
),
```

and the daemon reads that sidecar straight back into `last_scheduled_trigger_at` on load:

```rust
// src/routine_storage.rs
let last_scheduled_trigger_at = read_scheduled_state(dir_name);
// -> reads `last_scheduled_trigger_at` from scheduled.local.toml
```

So the sequence is:

1. Operator fires a routine on demand (UI button / REST / MCP). `svc_trigger` correctly stamps `last_manual_trigger_at` in `routine.toml`.
2. The spawned command *also* writes `last_scheduled_trigger_at = <now>` into `scheduled.local.toml`.
3. On the next store load the daemon ingests that value, and the routine now reports a `last_scheduled_trigger_at` for a run the cron schedule never produced.

This directly defeats the manual-vs-scheduled distinction that #128 (`last_triggered_at` only records manual triggers), #127 (runtime-state sidecar split) and #155 (`last_scheduled_trigger_at` to track cron-driven runs) were introduced to establish. Anything relying on `last_scheduled_trigger_at` — "did the schedule actually fire?", missed-run/catch-up logic (#462), next-run displays, health/observability — can be misled by an on-demand trigger.

## Motivation

The two timestamps exist precisely so an operator can tell a scheduled run apart from a manual one. A manual trigger silently overwriting the scheduled timestamp collapses that signal and makes `last_scheduled_trigger_at` untrustworthy.

## Acceptance criteria

- A manual trigger updates **only** `last_manual_trigger_at`; it must not change `last_scheduled_trigger_at` or the `scheduled.local.toml` sidecar.
- A real cron firing continues to stamp `last_scheduled_trigger_at` exactly as today.
- The scheduled-sidecar write is gated so it only runs on the scheduled (crontab) path — e.g. build the manual-trigger command without that statement, pass a flag through `build_routine_command`, or have the scheduled-vs-manual caller own the stamp — rather than baking it into the shared command builder.
- A regression test covers: invoking `svc_trigger` leaves `last_scheduled_trigger_at` unchanged (and `last_manual_trigger_at` updated), while the scheduled path still writes the sidecar.

---

tupe12334:
## Why

A manual trigger (`svc_trigger`, reached from the REST API, MCP `trigger_routine`, and the CLI) reuses `build_routine_command` — the very same launch script the OS crontab runs on schedule. That script unconditionally writes `last_scheduled_trigger_at` into the routine's gitignored `scheduled.local.toml` sidecar.

The daemon reads that sidecar back into `last_scheduled_trigger_at` on load, so an **on-demand** run was being recorded as if the **schedule** had fired. The UI/API "last scheduled fire" then reflected a manual run, conflating the two distinct signals (the issue's concern, follow-up to #128/#155).

The daemon already records the on-demand run separately as `last_manual_trigger_at` (set in `svc_trigger` before spawning), so the generated script must not also touch the scheduled-fire sidecar.

Fixes #478.

## What

- Add a `TriggerSource { Scheduled, Manual }` argument to `build_routine_command`. The sidecar-stamp statement is emitted **only** for `Scheduled`.
- `sync::routines` (the crontab/run.sh path) passes `Scheduled` — behaviour unchanged.
- `svc_trigger` (manual) passes `Manual` — no sidecar write.

## Tests

- New `build_routine_command_manual_omits_scheduled_trigger_sidecar`: a manual command contains no `last_scheduled_trigger_at` stamp, yet still builds the workbench and launches tmux.
- Existing `build_routine_command_stamps_scheduled_trigger_sidecar` keeps asserting the stamp is present for scheduled runs.

Local gate (rust 1.96.0): `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (522 passed) all green.

---
This pull request was opened by the "Daemon + Landing auto-improve PRs" routine of moadim. cc @tupe12334


tupe12334:
Blocked: superseded by upstream. `svc_trigger` (manual) and `svc_trigger_scheduled` (main) are already fully separate functions — `svc_trigger_scheduled` explicitly leaves `last_manual_trigger_at` untouched, and a manual trigger never touches the scheduled-fire sidecar. Duplicate concern of #874/#490. Needs a manual look before merge.

— PR comment commander

PR #542, opened 2026-06-21
fix(cron): guard cron-jobs sync against wiping the block on an empty store (#450)

### LINKED ISSUE #450: cron-jobs sync_to_crontab lacks the empty-store wipe guard that routines have — a failed/empty store load clobbers all managed job lines on the next mutation

tupe12334:
## Why

`sync_to_crontab` (cron-jobs forward sync) rebuilt and wrote the moadim crontab block **unconditionally**. When the in-memory `CronStore` is empty — which at sync time means the store **failed to load** (or a second daemon is racing this one), not a genuine "no jobs" state — the next job mutation wrote a bare block and **silently dropped every managed job's cron line**.

The routines forward sync (`sync_routines_to_crontab`) already guards against exactly this footgun. The cron-jobs path diverged and was missing the same protection. This closes #450.

## What

- `src/sync/mod.rs`: add the empty-store wipe guard to `sync_to_crontab` — if the store is empty **and** the live crontab still contains a `# moadim:<id>` job line, log a warning and skip the write, leaving existing lines intact. Introduces a `JOB_LINE_MARKER` const (`# moadim:`), distinct from the routines block's `# moadim-routine:` marker.
- `src/sync/cron_sync_tests.rs`: two tests mirroring the routines guard tests —
  - `sync_refuses_to_wipe_job_lines_when_store_is_empty` (guard fires, lines survive)
  - `sync_proceeds_when_store_empty_but_no_job_lines` (second operand false → sync proceeds)
- `CHANGELOG.md`: entry under `### Fixed`.

A store that loaded fine but holds only disabled/unmanaged jobs is **not** empty, so legitimately clearing the last job still works.

## Checks

- `cargo fmt --check` ✓
- `cargo clippy --all-targets` ✓ (no warnings)
- `cargo test sync::` ✓ (63 passed)

Closes #450

---


tupe12334:
Blocked: superseded by upstream. Duplicate concern of #460 — the empty-store wipe guard already exists in `src/sync/routines.rs`. Needs a manual look (likely just closing).

— PR comment commander

Issue #438, opened 2026-06-19
Expose per-routine live run status (is_running / active session) on GET /routines and the UI


tupe12334:
> **Suggestion** — filed by the *Suggest new GitHub issues (daemon + landing)* routine of moadim. Left **unassigned**; self-assign to approve.

## Problem

`GET /routines` (and the UI rows that render it) expose *static* and *scheduling* facts about a routine — `enabled`, `agent_registered`, `schedule_description`, `timezone`, `last_manual_trigger_at` — but **nothing about whether the routine is executing right now**. An operator cannot tell, from the API or the web UI, which routines currently have a live agent session versus which are idle. The only way to know is to drop to the host and `tmux ls`.

The daemon already has the primitive to answer this: `tmux_session_alive(session)` in `src/routines/cleanup/session.rs` does an exact-match `tmux has-session` probe. Today it is `pub(super)` and used only by the cleanup/watchdog sweep — the same liveness signal that overlap detection (#141) reasons about internally is never surfaced to callers.

## Motivation

- **Observability**: "is this routine running?" is the most basic operational question and currently unanswerable without shell access.
- **Composes with existing backlog**: pairs naturally with #369 (`next_run_at` — when it will *next* run) to give each routine row a full state: *running now* / *idle, next at HH:MM*. Distinct from #414 (aggregate Prometheus `active sessions`) and #95 (power-saving disable) — this is per-routine, live, in the primary REST/UI surface.
- **Low cost, high leverage**: the liveness check already exists; this is mostly plumbing it into the response model.

## Acceptance criteria

- [ ] Add a derived, non-persisted `is_running: bool` (name TBD) field to `RoutineResponse` in `src/routines/model.rs`, populated by probing the routine's expected tmux session(s) for a live agent.
- [ ] Promote / reuse the existing `tmux_session_alive` primitive (today `pub(super)` in `cleanup/session.rs`) rather than re-implementing the probe; keep the exact (`=`) target match so `moadim-foo-1` never matches `moadim-foo-10`.
- [ ] Account for the run-id session naming: a routine's live session uses its slugified title + run id, so the check must match any in-flight session for that routine (not a single fixed name).
- [ ] Surface the field on `GET /routines` and `GET /routines/{id}`, document it in the OpenAPI schema, and reflect it in the UI routine row (e.g. a "RUNNING" badge / indicator).
- [ ] A missing `tmux` binary is treated as "not running" (never an error), mirroring the existing best-effort behavior.
[...]

Issue #433, opened 2026-06-19
feat(notifications): alert on routine run failure via a configurable webhook/command hook


tupe12334:
## Problem

The daemon launches agent sessions on a schedule, but there is **no way to be told when a run fails**. Today a failure is silent:

- A routine whose agent exits non-zero just stops; the operator only finds out by manually opening the UI LOGS view or tailing `agent.log`.
- The max-runtime watchdog kills a hung tmux session, but reaping is invisible — no signal leaves the host.
- Many existing issues describe "silently no-ops / silent-failure" modes (e.g. #404, #383, #301, #187, #320). Even after each individual silent-failure is fixed, there is still no **active** channel that tells a human "routine X just failed."

The two existing observability issues do **not** cover this:
- #23 (logs from UI, run history, per-run log capture) is *passive* — you must go look.
- #414 (`/metrics` Prometheus endpoint) is *pull-based* — it requires an external scraper + alerting stack the single-host install doesn't have.

## Motivation

The product's whole point is unattended, scheduled agent runs. Unattended only works if failures page someone. A lightweight, zero-infra failure hook closes the loop without requiring Prometheus/Alertmanager, and directly mitigates the recurring silent-failure theme across the backlog.

## Proposal

Add a configurable **failure notification hook**, fired when a run ends abnormally (agent exits non-zero, or the watchdog kills it for exceeding `max_runtime_secs`).

- Configurable at global scope (and optionally per-routine override), e.g. a `[notifications]` block:
  - `on_failure_command = "..."` — run a shell command with run context in env vars (`MOADIM_ROUTINE`, `MOADIM_RUN_ID`, `MOADIM_EXIT_REASON`, `MOADIM_LOG_PATH`), **or**
  - `on_failure_webhook = "https://..."` — POST a small JSON payload (routine id/title, run id, started/ended, exit reason, tail of `agent.log`).
- Requires capturing the agent's terminal state/exit reason (exit code vs watchdog-kill) — coordinate with per-run capture work in #23.
- Opt-in: no config → no notifications, no behavior change.

## Acceptance criteria

- [ ] A documented config (global + optional per-routine) enables a failure command and/or webhook.
- [ ] On a run that exits non-zero, the hook fires exactly once with routine id/title, run id, exit reason, and a bounded `agent.log` tail.
- [ ] On a watchdog kill (max-runtime exceeded), the hook fires with a distinct `EXIT_REASON` (e.g. `timeout`).
[...]

Issue #414, opened 2026-06-19
feat(observability): expose a Prometheus /metrics endpoint (run counts, failures, durations, active sessions, disk)


tupe12334:
## Problem

The daemon is a long-running scheduler that launches agent sessions on cron/routine triggers, yet it exposes **no machine-readable operational metrics**. `GET /health` (`src/routes/http.rs`) is a static liveness stub — it only returns `status:"ok"`, `uptime_secs`, `running:true`, and `version`. There is no way to answer, without scraping logs or the UI, questions an operator running this as an OS service actually has:

- How many routine/cron runs have fired, succeeded, failed, or been reaped by the max-runtime watchdog?
- How many agent sessions are alive right now?
- What is the run-duration distribution?
- How much disk is the workbench tree consuming, and how many workbenches were reclaimed by cleanup?

Existing observability issues cover adjacent-but-different ground: UI log viewing / run history (#23), HTTP correlation ids (#354), and `daemon.log` rotation (#316). None expose a metrics surface a monitoring stack (Prometheus/Grafana, or a simple `curl | grep`) can scrape.

## Motivation

- The daemon ships an OS-service install (systemd/launchd). A `/metrics` endpoint lets operators alert on failed runs, stuck sessions, or disk growth instead of discovering them after the fact.
- Several silent-failure modes already tracked as issues (thundering herd #335, orphaned sessions #333/#320, disk exhaustion #398, zombie children #212) become *observable* once run/session/disk counters exist — metrics are the early-warning layer over those bugs.
- Cheap to scrape, format is stable text, no UI work required.

## Proposal

Add `GET /api/v1/metrics` returning Prometheus text exposition (`Content-Type: text/plain; version=0.0.4`). Instrument the existing run lifecycle (routine/cron trigger → spawn → finish/fail/reap) and cleanup sweep. No new heavy dependency is required — a tiny hand-rolled registry or a lightweight crate both fit.

Suggested initial series:

- `moadim_runs_total{kind="routine|cron",result="success|failure|reaped"}` — counter
- `moadim_active_sessions` — gauge (live tmux agent sessions)
- `moadim_run_duration_seconds` — histogram
- `moadim_workbench_bytes` — gauge (current workbench tree size; pairs with #398)
- `moadim_cleanup_removed_total` / `moadim_cleanup_freed_bytes_total` — counters (pairs with #311)
- `moadim_uptime_seconds`, `moadim_build_info{version,...}` — gauge/info

## Acceptance criteria

[...]

Issue #23, opened 2026-06-14
feat: observability — logs from UI, run history, per-run log capture


tupe12334:
## Overview

Three related features to make the moadim daemon observable from the UI and via MCP tools:

### 1. Per-run log capture (`feat/run-records` → PR #20)
Each time a managed job fires (scheduled or manual trigger), capture a structured `RunRecord`:
- run ID (UUID), job ID, start/end timestamps, wall-clock duration
- process exit code (`None` if handler couldn't be spawned)
- captured stdout + stderr (truncated to 10 KB each)
- trigger type (`scheduled` | `manual`)

Records stored in `~/.config/moadim/jobs/{id}/runs.jsonl` (newline-delimited JSON, newest-last, max 100 per job).

New REST endpoints:
- `GET /cron-jobs/{id}/runs` — list up to 100 most-recent runs, newest first
- `GET /cron-jobs/{id}/log` — return raw `job.log` contents as `{content: string}`

New MCP tools:
- `get_job_runs` — list execution history for a cron job
- `get_job_log` — get the raw log file

### 2. Run history modal in UI (`feat/ui-runs` → PR #21)
- **LOGS** button added to each job row
- Opens a modal showing up to 100 most-recent runs
- Per-row: index, relative start time, duration, exit-code badge (OK / EXIT N / FAILED), trigger type
- Expand row to see captured stdout + stderr

### 3. Raw log tab in UI (`feat/ui-log` → PR #22)
- Extends the LOGS modal with a **RUNS | LOG** tab switcher
- **LOG tab** renders the raw append-only `job.log` in a scrollable `<pre>` block
- Both endpoints fetched on modal open

## PR chain
`fix/execute-managed-jobs` ← `feat/run-records` (#20) ← `feat/ui-runs` (#21) ← `feat/ui-log` (#22)

PR #543, opened 2026-06-21
feat(routines): record each run's terminal exit status (#453)

### LINKED ISSUE #453: Routine runs record no exit status — a crashed/errored run is indistinguishable from a successful one
## Problem

A routine run is launched as a detached tmux session whose pane runs the agent directly, with its stdout/stderr piped to `agent.log`:

```
tmux new-session -d -s "$SESS" -c "$WB" <agent invocation>
tmux pipe-pane -o -t "$SESS" 'cat >> "$WB"/agent.log'
```

When the agent process exits, the tmux session simply ends. **Its exit code is never captured anywhere.** As a result, after a run finishes, three very different outcomes are indistinguishable to the daemon, the API, the UI, and an operator:

- the agent completed successfully (exit `0`),
- the agent exited with an error (non-zero, e.g. a crash, an auth failure, a panic), and
- the run was force-killed by the max-runtime watchdog (the only outcome that *is* recorded today, via the `moadim: routine exceeded max runtime; killing session` note appended to `agent.log` in `note_forced_kill`).

The cleanup sweep reaps a finished workbench purely on "no live tmux session + TTL elapsed" — it has no notion of whether the run *worked*. So a routine can fail on every single tick (bad token, broken prompt, agent binary erroring out) while every signal the system exposes — green health dot, normal reaping, a populated `agent.log` — looks identical to success.

## Motivation

Run outcome is a foundational signal that several already-filed features need but none of them provide, because each *consumes* a failure/success signal that does not yet exist:

- **#433** (alert on routine run failure via webhook/command) has nothing to fire on — there is no "this run failed" event to hook.
- **#414** (Prometheus `/metrics`: run counts, **failures**, durations) cannot count failures without a per-run status.
- **#23 / #438** (run history / live run status) can show that a run *ran* but not whether it *succeeded*.

This issue is the missing primitive underneath all of those: **persist each run's terminal exit status**, so finished-but-failed is distinguishable from finished-and-succeeded. It deliberately does not implement alerting, metrics, history UI, or live-status — only the capture mechanism those build on.

A natural, crontab-line-friendly approach: wrap the agent invocation so its exit code is recorded into the workbench when it ends, e.g. run `<agent>; echo $? > "$WB/exit_code"` inside the tmux pane (or use tmux `remain-on-exit` + `pane_dead_status`), and have the watchdog-kill path write a distinct sentinel (e.g. `killed`) so the three outcomes are separable. The cleanup sweep and any future status surface can then read this file before the workbench is reaped.

## Acceptance criteria

- [ ] Every routine run writes its terminal status to a well-known, per-run location in the workbench (e.g. `$WB/exit_code` / a small status file) when the agent process exits — for both scheduled (crontab) and manual (`trigger`) launches, which share `build_routine_command`.
- [ ] The three outcomes are distinguishable: success (`0`), agent error (non-zero exit code preserved), and watchdog-killed (a distinct sentinel, not a misleading `0`/blank).
- [ ] The capture survives until the workbench is reaped (it is written before the session ends, not racing the reap), and adds no new hard dependency or unbounded growth.
- [ ] Status capture is covered by tests in the same pure/injected style as the existing cleanup and command builders (no reliance on a live tmux server or filesystem clock).
- [ ] Docs (Architecture.md) note where per-run status is recorded and what the sentinels mean.
- [ ] Non-goals (explicitly out of scope here, to avoid overlap): the webhook/alert hook (#433), the Prometheus endpoint (#414), the run-history/log UI (#23), and live in-progress status (#438). This issue only lands the underlying captured-status signal they each need.

---

*This issue was suggested (unassigned) by the "Suggest new GitHub issues (daemon + landing)" routine of moadim — a suggestion for @tupe12334 to review and self-assign if useful.*

tupe12334:
## Problem

A routine run is launched as a detached tmux session whose pane runs the agent directly. When the agent exits, the tmux session simply ends — **its exit code is never captured anywhere**. As a result a finished run's three very different outcomes were indistinguishable to the daemon, API, UI, and an operator:

- the agent completed successfully (`0`),
- the agent exited with an error (non-zero — crash, auth failure, panic),
- the run was force-killed by the max-runtime watchdog.

Every signal — green health dot, normal reaping, a populated `agent.log` — looked identical to success, so a routine could fail on every tick while appearing healthy.

## Change

Persist each run's terminal status to a well-known per-run file, `$WB/exit_code`:

- **Capture (both launch paths):** `build_routine_command` suffixes the agent invocation with `; echo $? > exit_code`. The pane's cwd is the workbench (`tmux -c "$WB"`), so the relative file lands in the run's workbench. The scheduled (crontab) and manual (`trigger`) paths both build the command here, so both are covered.
- **Watchdog sentinel:** `note_forced_kill` writes the distinct `killed` sentinel to `exit_code` instead of letting a force-killed run masquerade as a clean `0`. The kill SIGKILLs the pane before its own `echo` can run, so there is no clobber.

The three outcomes are now distinguishable: `0` (success), a preserved non-zero (agent error), and `killed` (watchdog). The file is written before the session ends (not racing the reap) and survives until the workbench is reaped under the normal `ttl_secs` rules. No new dependency, no unbounded growth.

### Scope

This lands **only** the captured-status primitive. The consumers that each need it remain explicitly out of scope: failure alerts (#433), `/metrics` failure counts (#414), run history / live status (#23 / #438).

## Tests & docs

- `build_routine_command_captures_exit_status` asserts the capture is part of the tmux-launched invocation.
- Updated `build_routine_command_substitutes_arg_placeholders` for the new wrapped invocation.
- Extended the watchdog reap test to assert `exit_code == "killed"`.
- `Architecture.md` documents the `exit_code` location and the sentinel meanings.
- `CHANGELOG.md` updated under `[Unreleased] → Added`.

Local checks pass: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` (522 passed).

Closes #453

---
*This pull request was opened by the "Open issues → fix PRs (owned repos + my orgs)" routine of moadim.*


tupe12334:
Blocked: superseded by upstream. Each run's terminal exit status is already recorded — `RunStatus`/`exit_code` in `src/routines/model.rs` (main), populated by the run-history feature. Needs a manual look (likely just closing).

— PR comment commander

Issue #150, opened 2026-06-17
CI runs only fmt + clippy — tests and the 100% coverage gate are enforced only by local hooks


tupe12334:
## Problem

CI enforces only **formatting** and **lint**, not the test suite or the coverage gate.

- `.github/workflows/lint.yml` runs `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings`.
- The other workflows (`changelog.yml`, `spellcheck.yml`, `publish.yml`) do not run `cargo test` either — `grep 'cargo test\|llvm-cov' .github/workflows/*.yml` returns nothing.
- Meanwhile `CONTRIBUTING.md` documents that the **pre-push hook** runs both `cargo test` and `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'`.

So the repo's two hardest quality gates — the ~23 test modules and the 100% line-coverage floor — are enforced **only locally**, and only for contributors who ran `git config core.hooksPath .githooks`. A PR from a fork, or from anyone who skipped hook setup, can break tests or drop coverage and still go green.

This is the exact failure mode `lint.yml`'s own header comment calls out for fmt/clippy ("caught on every PR even when a contributor hasn't installed the git hooks") — but tests and coverage get no such CI net.

## Motivation

- `main` can regress on tests/coverage silently between local hook runs — the same drift the fmt/clippy CI gate was added to stop.
- The 100% coverage promise in `CONTRIBUTING.md` is unverifiable on the server side; it is documentation, not enforcement.
- Fork PRs never run the local hooks at all, so external contributions are entirely ungated on correctness.

## Acceptance criteria

- A CI job runs `cargo test` on every `pull_request` and on push to `main`.
- A CI job (same or separate) runs `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'`, mirroring the pre-push hook, and fails the build below 100%.
- The job uses the committed `prebuilt.html` fallback (no trunk/wasm needed), consistent with `lint.yml`.
- Coverage/test commands stay in lockstep with `CONTRIBUTING.md` (single source of truth, or a comment cross-referencing it) so the documented gate and the CI gate can't drift.

---
_This issue was opened by the "Suggest new GitHub issues (daemon + landing)" routine of moadim, as a suggestion. It is intentionally left unassigned — self-assign to approve it._

tupe12334:
[...]

PR #548, opened 2026-06-21
fix(sync): order tied-created_at routines by id in the crontab block

tupe12334:
## Why

`build_block` (`src/sync/routines.rs`) collects enabled managed routines from a `HashMap` and sorted them **only by `created_at`**:

```rust
routines.sort_by_key(|routine| routine.created_at);
```

`created_at` is not unique. Routines that share a `created_at` — e.g. several seeded by `ensure_default_routines` or batch-created by a script in the same second — therefore emit in an **arbitrary, run-to-run order**, since `HashMap` iteration order is unspecified.

This churns the generated routines crontab block across syncs and defeats the `new_crontab == current` idempotency guard in `sync_routines_to_crontab`:

```rust
if new_crontab == current {
    return Ok(());
}
write_crontab(&new_crontab)
```

When tied routines reshuffle, `new_crontab != current` spuriously, so every sync/startup fires a needless `crontab -` rewrite — a subprocess that mutates the user's **live** crontab and can race a concurrent daemon — even though nothing actually changed.

## What

Break ties on the stable routine `id`:

```rust
routines.sort_by(|left, right| {
    left.created_at
        .cmp(&right.created_at)
        .then_with(|| left.id.cmp(&right.id))
});
```

This mirrors the same determinism fix already applied to `svc_list` for the `GET /routines` listing, keeping the two sort paths consistent.

Adds `build_block_orders_tied_created_at_by_id_deterministically`: inserts two tied routines in id-descending order and asserts they emit in id order and that repeated builds are byte-identical.

## Verification

- `cargo fmt --check` — clean
- `cargo test sync::routines` — 13 passed
- `cargo llvm-cov` — the new sort closure and test lines are fully covered; this change adds no uncovered lines (the tree's pre-existing sub-100% coverage is unchanged and tracked separately under #150).

---


tupe12334:
LGTM

tupe12334:

tupe12334:
Heads up: this PR is approved (LGTM) but currently **cannot be merged** — the branch is in a CONFLICTING state with `main`. Please rebase / resolve the merge conflicts and it will be ready to merge.

PR #550, opened 2026-06-21
feat(cli): surface server uptime_secs in `status --json`

tupe12334:
## What

`moadim status --json` now includes an `uptime_secs` field carrying the running server's age, so a single call returns running-state, local pid, bound address, **and** server age:

```json
{"running":true,"pid":4821,"address":"127.0.0.1:5784","uptime_secs":3600}
```

`uptime_secs` is `null` when the server is down or its `/health` body carries no parseable age (mirroring how `pid` is `null` when unknown).

## Why

The server already reports `uptime_secs` on `GET /health`, but `status --json` only surfaced liveness + local pid. A script wanting both how-long-up and is-it-up had to make a second call and parse `/health` itself. This folds the age into the existing liveness probe — `status` already hits `GET /health` to decide `running`, so reading the body it gets back for free is one probe, not two. (Implements the standing `status --json` enrichment item in `TODO.md`.)

## How

- `status` now calls a single `probe_health()` that returns `(running, uptime_secs)` from one `GET /health`.
- `parse_uptime_secs` extracts the field defensively — non-JSON, a missing field, or a non-numeric value all degrade to `None` rather than erroring.
- The human-readable `status` line and the `stop` / `cleanup` JSON shapes are unchanged. `stop --json` still shares `status`'s `running` / `pid` / `address` fields (it has no server to report an age for); README wording adjusted to say "fields match" rather than "same shape".

## Tests / gates

- New unit tests cover the parser's present / absent / non-numeric / non-JSON paths and the enriched object.
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo llvm-cov --fail-under-lines 100` all pass locally (100% line coverage retained).
- README and CHANGELOG (`### Added`) updated.

---


tupe12334:
Blocked: superseded by upstream. `status --json` on `main` already folds in server-sourced `uptime_secs` (`src/cli.rs`). Needs a manual look before merge.

— PR comment commander

PR #553, opened 2026-06-21
fix(uninstall): clear managed crontab blocks on uninstall (#380)

### LINKED ISSUE #380: moadim uninstall removes the OS service but leaves the managed crontab blocks — routines keep firing (or erroring) after uninstall

## Problem

`moadim uninstall` only removes the OS service registration — it never clears the managed crontab blocks that drive routine execution.

- `src/service/macos.rs::uninstall()` unloads + deletes the launchd plist.
- `src/service/linux.rs::uninstall()` disables + deletes the systemd user unit.
- `src/cli.rs` maps `Some("uninstall") => Command::Uninstall`, which calls only `service::uninstall()`.

Neither path touches the user's crontab. But routines are scheduled as real OS crontab entries inside the managed blocks (`src/sync/routines.rs`: `# BEGIN MOADIM-ROUTINES` … `# END MOADIM-ROUTINES`, and `src/sync/mod.rs`: `# BEGIN MOADIM` … `# END MOADIM` for cron-jobs).

So after `moadim uninstall` the user's crontab still contains every managed line, and `cron` keeps firing them on schedule. Each tick re-invokes the routine's `run.sh` wrapper (tmux + agent launch), so a daemon the user explicitly **uninstalled** keeps spawning agent sessions — or, if the binary/wrapper was also removed, every interval errors and cron mails the failure (see also #375). Uninstall is not a clean teardown; the user is left to hand-edit `crontab -e` and delete the `# BEGIN MOADIM*` blocks themselves.

## Motivation

- Principle of least surprise: "uninstall" should remove what install/usage added. Install registers a service **and** sync writes crontab blocks; uninstall removes only the former.
- A tool that mutates the user's crontab owns cleaning it up. Orphaned cron lines invoking a stopped/removed daemon are the worst kind of residue — invisible until they fire, then either silently act (open PRs/issues) or spam cron mail forever.
- The block-replace plumbing already exists: `replace_block_with(&current, "", BLOCK_BEGIN, BLOCK_END)` removes a managed block, and the empty-store path in `sync_routines_to_crontab` / `sync_to_crontab` already proves a block can be cleared.

## Acceptance criteria

- [ ] `moadim uninstall` removes BOTH managed crontab blocks (`# BEGIN MOADIM-ROUTINES`/`END` and `# BEGIN MOADIM`/`END`) in addition to removing the OS service, leaving the rest of the user's crontab untouched.
- [ ] Removal is idempotent and tolerant: a crontab with no managed block, no crontab at all, or `crontab` not on PATH must not make uninstall fail (mirror the best-effort style already used for `launchctl unload` / `systemctl disable`).
- [ ] The unsupported-platform `uninstall()` stub (`src/service/mod.rs`) is reconciled — crontab cleanup should still run on platforms where the service step is unsupported but crontab sync is used.
- [ ] User-facing output reports what was cleared (e.g. "removed N managed crontab entries"), consistent with the existing service-removal messages.
- [ ] Test coverage: uninstall against a sandboxed crontab (reuse the `CRONTAB_BIN`/`PATH` isolation seam from #175) asserts the managed blocks are gone and surrounding user entries are preserved; a second uninstall exercises the already-clean branch without error.

## Out of scope / related

- Killing in-flight agent sessions at uninstall time — tracked separately (#320, #333).
- Marker-collision robustness between the two block names — #324.

tupe12334:
## Summary

`moadim uninstall` only removed the OS service registration — it never cleared the managed crontab blocks that actually drive routine/job execution. After uninstall the user's crontab still held every managed line, so `cron` kept firing them: either silently re-spawning agent sessions against a daemon the user removed, or (once the wrapper/binary was gone) erroring every interval and spamming cron mail. Uninstall was not a clean teardown.

## Change

- Add `sync::clear_managed_crontab_blocks()`:
  - Removes **both** managed blocks — `# BEGIN MOADIM-ROUTINES`/`END` and `# BEGIN MOADIM`/`END` — leaving every other crontab entry untouched.
  - Clears the **routines block first**: its begin marker is a superstring of the cron-jobs marker, so doing cron-jobs first could match the routines marker by substring and wipe the wrong block (the #324 collision). Each removal is gated on the block being present, so the no-op path never rewrites an otherwise-clean crontab (idempotent).
  - Returns the number of managed schedule lines removed, for a user-facing report.
  - Best-effort: no managed block, no crontab at all, or `crontab` not on PATH all succeed without error.
- Wire it into the `uninstall` command next to `service::uninstall()`. The two steps are **independent**: a failed or unsupported-platform service step is reported but does not skip the crontab cleanup, and the command still succeeds so a partial install can always be torn down.
- Reconcile the unsupported-platform path (crontab cleanup runs regardless of service support).
- Update the CLI help text and CHANGELOG.

## Verification

New tests (reusing the `MOADIM_CRONTAB_BIN` sandbox seam) assert: both blocks removed with surrounding user entries preserved; a second uninstall hits the already-clean branch returning 0 with the crontab byte-for-byte unchanged; a crontab with no managed block is untouched; an absent crontab succeeds. `cargo fmt`, `cargo clippy`, and `cargo test` (525 passed) all green locally.

Closes #380

---
This pull request was opened by the *Open issues → fix PRs (owned repos + my orgs)* routine of moadim. cc @tupe12334


tupe12334:
LGTM

tupe12334:

tupe12334:
Heads up: I attempted to merge this after your LGTM, but the branch currently has merge conflicts with the base (mergeable state: DIRTY), so it cannot be merged automatically. Please resolve the conflicts (rebase/merge the base branch) and I will merge on the next run, or merge it yourself.

PR #563, opened 2026-06-21
test(cleanup): restore 100% line coverage so the pre-push gate passes on stable

tupe12334:
## Why

The pre-push hook (and the gate documented in `CONTRIBUTING.md`) runs:

```
cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'
```

On a clean `main` checkout that command **fails** — coverage sat at **99.90%** (2–3 missed lines), so nobody can `git push` without `--no-verify`. This restores the 100% line floor with test-only changes.

## What was uncovered

All three gaps were in the cleanup module, and all are the hard-to-reach "a real tmux is present" paths that the existing fake-injected unit tests deliberately avoid:

- **`session::tmux_bin`** — the `MOADIM_TMUX_BIN` override branch. The existing test only *removes* the var to verify the `cfg(test)` fallback, so the line that returns the override never ran.
- **`session::tmux_session_alive`** — the `|status| status.success()` mapping. It only executes when the spawned process returns a status; with no tmux installed `Command::status()` always errors and the closure is skipped via `unwrap_or(false)`.
- **`cleanup::cleanup_expired_workbenches`** — the `max_runtime_for` closure, which is consulted only on the watchdog (live-session) path. The existing end-to-end test runs without tmux, so the session is never "alive" and that path is never taken.

## What this adds

Two tests in `cleanup_tests.rs` (no production code touched):

1. `tmux_bin_honors_the_env_override` — asserts `tmux_bin()` returns the `MOADIM_TMUX_BIN` value verbatim, complementing the existing no-override case.
2. `cleanup_expired_workbenches_kills_a_live_hung_session` (`#[cfg(unix)]`) — points `MOADIM_TMUX_BIN` at a stub script that always exits 0, so a session reports **alive**, the watchdog trips on an ancient timestamp, the session is force-killed, and the workbench is reaped. This exercises the `status.success()` mapping, the `max_runtime_for` closure, and the real kill/`note_forced_kill` side-effects in one pass.

Both tests restore/save the env vars they touch and rely on the crate's single-threaded test harness, matching the surrounding tests.

## Verification

```
cargo fmt --check                                                   # clean
cargo clippy --all-targets -- -D warnings                          # clean
cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'   # PASS — lines & functions 100%
```

Note: this closes the *coverage value* gap; making CI actually enforce the gate is the separate concern tracked in #150.

---
This pull request was opened by the "Daemon + Landing auto-improve PRs" routine of moadim.


tupe12334:
LGTM

tupe12334:
Heads up: I went to merge this after your LGTM, but the branch currently has merge conflicts with the base (mergeable state: DIRTY/CONFLICTING), so it cannot be merged automatically. Please resolve the conflicts (rebase/merge the base branch) and I will merge on the next run, or merge it yourself.

tupe12334:

PR #609, opened 2026-06-21
fix(routines): anchor workbench under workbenches_dir() so launch honors MOADIM_HOME_OVERRIDE (#601)

### LINKED ISSUE #601: Workbench launch path is hardcoded ($HOME/.moadim/workbenches) instead of derived from paths::workbenches_dir() — drifts from the reaper/LOGS view and ignores MOADIM_HOME_OVERRIDE (#585 follow-up)
## Problem

The workbench base path `~/.moadim/workbenches/{slug}-{ts}` is defined in **two independent places** that can silently drift:

1. **Source of truth (Rust):** `paths::workbenches_dir()` → `moadim_home().join("workbenches")`, where `moadim_home()` resolves its base through `paths::home()`, which **honors the `MOADIM_HOME_OVERRIDE` seam** (added in #585). Every consumer that *reads or reaps* a workbench uses this:
   - `routines/cleanup/mod.rs:123` — the TTL reaper sweep
   - `routines/service.rs:473` — listing runs for the LOGS view
   - `routines/service.rs:493` — reading a run's `agent.log`

2. **Where runs are actually written (shell):** `routines/command.rs:230` hardcodes the literal
   ```sh
   ```
   This uses the raw shell `$HOME` and a hand-written `.moadim/workbenches` string — it does **not** go through `paths::workbenches_dir()` and is **blind to `MOADIM_HOME_OVERRIDE`**.

So the daemon *creates* workbenches at one path and *cleans up / reads logs from* a path computed a different way. They only coincide because today `paths::home()` falls back to `dirs::home_dir()` (≈ `$HOME`) and the layout literal happens to match.

## Motivation

The two diverge the moment either side changes:

- **`MOADIM_HOME_OVERRIDE` set on the daemon process:** the reaper and LOGS view scan `$OVERRIDE/.moadim/workbenches`, but the cron launch writes to `$HOME/.moadim/workbenches`. Result: **workbenches leak forever** (the reaper never sees them → compounds the unbounded-growth concerns of #398/#430) and the **LOGS view shows nothing** for real runs. #585 isolated the daemon's Rust paths behind this seam but left the generated shell command outside it.
- **Any future layout change** (e.g. honoring XDG #328, relocating the home base, renaming the dir): editing `paths/mod.rs` looks complete and compiles green, but the hardcoded shell string keeps writing to the old location — a classic two-sources-of-truth drift with no compile-time or test guard. The man-page drift guard in #556 is the same class of problem.

This is a maintainability/correctness gap, not a user-facing feature: the single source of truth (`paths::workbenches_dir()`) should drive the launch command too.

## Acceptance criteria

- [ ] `build_routine_command` derives the workbench base directory from `paths::workbenches_dir()` (or `moadim_home()`) rather than the hardcoded `$HOME/.moadim/workbenches` literal — e.g. the resolved absolute base is shell-quoted and interpolated, with `$SLUG-$TS` appended at run time.
- [ ] The override seam is respected: with `MOADIM_HOME_OVERRIDE` set, the path a run is created at matches the path the reaper (`cleanup/mod.rs`) and the LOGS view (`service.rs`) scan.
- [ ] A regression test asserts that the workbench dir in the generated launch command shares a base with `paths::workbenches_dir()` (so the two can't silently drift again) — ideally exercised under a `MOADIM_HOME_OVERRIDE` tempdir.
- [ ] No behavior change for the default install (no override set): the path remains `~/.moadim/workbenches/{slug}-{ts}`.

---
_This issue was filed as a non-binding suggestion by the "Suggest new GitHub issues (daemon + landing)" routine of moadim. It is intentionally left **unassigned**; @tupe12334 can approve it by self-assigning. Please re-check for duplicates before acting._

tupe12334:
## Why

`build_routine_command` hardcoded the workbench root as:

```sh
```

Every other consumer — the reaper (`svc_logs`, cleanup) and the LOGS view — derives directories from `paths::workbenches_dir()`, which honours `MOADIM_HOME_OVERRIDE`. So a daemon run with the override set launched routines under `$HOME/.moadim/...` while the rest of the daemon looked under the override path: the launch path drifted from the reaper/LOGS view, and finished runs were never reaped from the location they actually ran in. Closes #601.

## What

Bake the daemon-resolved `workbenches_dir()` into the generated launch script instead of hardcoding `$HOME`. The resolved root is single-quoted (a shell literal) and placed adjacent to the double-quoted `"$SLUG-$TS"`, so the shell concatenates the two while still expanding the run-time slug and timestamp:

```sh
WB='/resolved/workbenches'/"$SLUG-$TS"
```

This mirrors the existing `user_prompt_path().to_string_lossy()` resolution a few lines below. When no override is set, `workbenches_dir()` resolves to `~/.moadim/workbenches`, so behaviour is unchanged for the default case.

## Tests

Adds `build_routine_command_anchors_workbench_under_workbenches_dir`, which sets `MOADIM_HOME_OVERRIDE` to a temp home and asserts the override-anchored root is baked in and that `$HOME/.moadim/workbenches` is no longer hardcoded. `cargo fmt --check`, `cargo clippy --all-targets`, and the command-module tests all pass locally. CHANGELOG updated under `## [Unreleased] → Fixed`.

---
This pull request was opened by the **Daemon + Landing auto-improve & auto-merge lint/docs PRs** routine of moadim. cc @tupe12334


tupe12334:
Blocked: superseded by upstream. `build_routine_command` in `src/routines/command.rs` (main) already derives the workbench launch path from `crate::paths::workbenches_dir()` (which honors `MOADIM_HOME_OVERRIDE`), not a hardcoded `$HOME/.moadim/workbenches`. Duplicate concern of #610. Needs a manual look before merge.

— PR comment commander

Issue #585, opened 2026-06-21
Unit tests write into the real ~/.config/moadim home (missing MOADIM_HOME_OVERRIDE isolation)


tupe12334:
## Summary

Several unit tests write into the developer's **real** home directory (`~/.config/moadim/routines/…`) instead of an isolated temp home. This pollutes real state, races other tests, leaks on panic, and in the worst cases the migration tests **scan and mutate the real routines directory**.

## Root cause

`crate::paths::home()` honors the `MOADIM_HOME_OVERRIDE` test seam, but falls back to the **real** home when the env var is unset:

```rust
pub(crate) fn home() -> Option<PathBuf> {
    match std::env::var_os("MOADIM_HOME_OVERRIDE") { /* … else real home */ }
}
```

Any test that calls `write_routine` / `svc_create` / `svc_update` / `migrate_routine_dirs` **without** setting the override therefore resolves `routine_dir(...)` under the real `~/.config/moadim`.

## Affected tests

**`src/routines/service_tests.rs`** — *none* of the 45 tests set the override. Success-path tests leak slugs to real home:
- `valid_create_request()` → title `"Valid Title"` → slug `valid-title` (used by `svc_create_persists_machines`, `svc_create_accepts_secs_at_cron_ceiling`, `svc_create_syncs_crontab_on_success`, `svc_create_accepts_builtin_agent`, `svc_create_trims_repository_entries`, …)
- `svc_update_rejects_renaming_into_existing_slug` → title `"Keep"` → slug `keep`

[...]

PR #610, opened 2026-06-21
fix(routines): derive workbench launch path from paths::workbenches_dir() (#601)

### LINKED ISSUE #601: Workbench launch path is hardcoded ($HOME/.moadim/workbenches) instead of derived from paths::workbenches_dir() — drifts from the reaper/LOGS view and ignores MOADIM_HOME_OVERRIDE (#585 follow-up)

tupe12334:
## Problem

`build_routine_command` hardcoded the workbench base as the shell literal `$HOME/.moadim/workbenches/$SLUG-$TS` — a **second source of truth**. Every consumer that *reads or reaps* a workbench instead resolves the base through `paths::workbenches_dir()`, which honors the `MOADIM_HOME_OVERRIDE` seam added in #585:

- `routines/cleanup/mod.rs` — the TTL reaper sweep
- `routines/service.rs` — the LOGS view (listing runs, reading `agent.log`)

With `MOADIM_HOME_OVERRIDE` set, runs were **created** at `$HOME/.moadim/workbenches` while the reaper and LOGS view **scanned** `$OVERRIDE/.moadim/workbenches`: workbenches leak forever (the reaper never sees them) and the LOGS view shows nothing for real runs.

## Fix

Resolve the base once from `paths::workbenches_dir()`, shell-quote it, and append `/$SLUG-$TS` at run time:

```rust
let workbenches_base = crate::paths::workbenches_dir().to_string_lossy().into_owned();
...
format!(r#"WB={}/"$SLUG-$TS""#, shell_quote(&workbenches_base)),
```

- **Default install unchanged**: no override → base stays `~/.moadim/workbenches`, so runs still land at `~/.moadim/workbenches/{slug}-{ts}`.
- **Override now flows to the launch command**, so the create-side and the read/reap-side can no longer silently drift.

## Tests

Adds `build_routine_command_workbench_path_honors_home_override`: with a `MOADIM_HOME_OVERRIDE` tempdir set, it asserts the generated `WB=` line is anchored at exactly the `paths::workbenches_dir()` base.

`cargo test`, `cargo clippy --all-targets -- -D warnings`, and `cargo fmt --check` all pass locally.

Closes #601

---
🤖 This pull request was opened by the "Open issues → fix PRs (owned repos + my orgs)" routine of moadim. cc @tupe12334


tupe12334:
Blocked: superseded by upstream. Same fix as #609 — `build_routine_command` already anchors the workbench path via `crate::paths::workbenches_dir()`. Needs a manual look before merge.

— PR comment commander

PR #619, opened 2026-06-21
fix(machines): validate routine/job machines list on create & update (#600)

### LINKED ISSUE #600: machines targeting list is not validated — empty/whitespace entries silently exclude a routine/job (and an all-empty list evades the dormant warning)
## Problem

The `machines` targeting list on routines and cron jobs is stored **verbatim** on create/update — no trim, no empty-filter, no dedup — while every sibling input is validated server-side:

- `title` → `validate_title` (#226)
- `repositories` → `validate_repositories` (#241)
- cron `handler` → validated (#252)

In `src/routines/service.rs` the create path runs `machines: req.machines` and the update path runs `routine.machines = machines` directly (same in `src/cron_jobs.rs`). There is no `validate_machines` helper.

Matching is **exact** in `src/machine/mod.rs`:

```rust
pub fn targets(machines: &[String], me: &str) -> bool {
    machines.iter().any(|name| name == me)
}
```

So any entry that isn't byte-for-byte equal to the resolved machine name never matches. Two silent-failure modes result:

1. **Whitespace / typo entry** (`" my-laptop"`, trailing `\n`) — never equals `me`, so the routine/job is silently excluded from this machine's crontab even though the operator believes it's targeted here.
2. **Empty-string entry** (`[""]`, or `["", "other"]`) — the list is *non-empty*, so it slips past the dormant-routine warning (`warn_dormant_jobs` / the routines equivalent only fire on `machines.is_empty()`), yet `targets` can never match `""`. Result: a routine that **runs nowhere with no warning at all** — strictly worse than the empty-list case, which at least logs a dormant notice.

The web UI's picker already trims and rejects blanks (`ui/src/machines.rs` `add_new`), but the REST API, the MCP create/update tools, a hand-edited `routine.toml`, and a `machines` list authored on another host in the shared config repo all reach the store unguarded. Server-side validation is the authoritative seam — which is exactly why `repositories`/`title`/`handler` are validated on the server despite the UI also guarding them.

## Motivation

- Closes a silent-failure gap consistent with the repo's established "reject empty/whitespace input on create/update" pattern (#226 / #241 / #252).
- The empty-string-entry case is a genuine correctness hole: it defeats the dormant-machine warning that exists specifically to stop a routine from silently never firing.
- The `machines` feature is new (0.15.0) and multi-machine config is shared via git across hosts, so malformed entries are easy to introduce by hand and hard to debug (green health dot, never fires).

## Acceptance criteria

- [ ] Add a `validate_machines(&[String]) -> Result<Vec<String>, AppError>` helper (mirroring `validate_repositories`) used by both routine and cron-job create **and** update paths.
- [ ] Reject any entry that is empty or whitespace-only with a clear, indexed error (e.g. `machines[1] must not be empty or whitespace-only`).
- [ ] Trim each accepted entry, and dedupe the list, before persisting (so `" host "` and `"host"` can't both linger and one of them silently never matches).
- [ ] Validation applies on `POST`/`PUT` for both routines and cron jobs, across REST and MCP (covered by the shared service layer).
- [ ] Unit tests: whitespace-only entry rejected; empty-string entry rejected; surrounding whitespace trimmed; duplicates collapsed; a valid list round-trips unchanged.
- [ ] (Optional) Note in the error or docs that matching is exact and case-sensitive, so operators know names must match `moadim machine show` exactly.

---

_Suggested by the **Suggest new GitHub issues (daemon + landing)** routine of moadim — filed UNASSIGNED as a suggestion. Self-assign to approve. @tupe12334_

tupe12334:
## What

Adds a shared `validate_machines(&[String]) -> Result<Vec<String>, AppError>` helper in `src/machine/mod.rs` (next to `targets`) and wires it into **both** the routine and cron-job **create and update** service paths (`src/routines/service.rs`, `src/cron_jobs.rs`). It:

- rejects any entry that is empty or whitespace-only with a clear, indexed error (`machines[1] must not be empty or whitespace-only`);
- trims each accepted entry;
- dedupes the list (first occurrence wins, order preserved) before persisting.

## Why

The `machines` targeting list was stored **verbatim** on create/update — no trim, no empty-filter, no dedup — while every sibling input (`title`, `repositories`, cron `handler`) is validated server-side. Matching in `machine::targets` is exact (`name == me`), so:

1. a whitespace/typo entry never matches → the routine/job is silently excluded from this machine's crontab; and
2. an empty-string entry keeps the list non-empty, so it slips past the dormant-routine warning (which only fires on `machines.is_empty()`) yet can never match — a routine that **runs nowhere with no warning at all**.

Server-side validation is the authoritative seam (REST, MCP, and hand-edited `routine.toml`/`job.toml` all reach the store), mirroring the existing `validate_repositories`/`validate_title` guards.

## Tests

Six unit tests in `src/machine/mod_tests.rs`: clean list round-trips, surrounding whitespace trimmed, duplicates collapsed after trimming, empty-string entry rejected with an indexed error, whitespace-only entry rejected, empty list accepted unchanged.

`cargo build`, `cargo clippy --all-targets`, `cargo fmt --check`, and `cargo test machine::` all pass locally.

Closes #600

---
🤖 This pull request was opened by the **Open issues → fix PRs (owned repos + my orgs)** routine of moadim.


tupe12334:
Blocked: superseded by upstream. `validate_machines` in `src/routines/service.rs` already trims, rejects empty/whitespace entries with an indexed error, and dedupes — wired into both create and update. Needs a manual look (likely just closing).

— PR comment commander

PR #650, opened 2026-06-24
fix(trigger): reap spawned child processes to prevent zombies

### LINKED ISSUE #212: Manual routine/cron-job triggers leak zombie (<defunct>) child processes — spawned Child is dropped without wait()

tupe12334:
## Summary

Both manual-trigger paths (`svc_trigger` in `src/routines/service.rs` and `svc_trigger` in `src/cron_jobs.rs`) were dropping the `Child` handle immediately after `.spawn()`. Rust does not reap children on `Child` drop, so each successful trigger leaked a zombie (`<defunct>`) process that lived until the daemon exited.

- In `spawn_routine_command` (`src/routines/service.rs`): change `if let Err(...)` on `.spawn()` to a `match`, capturing `Ok(mut child)` and spawning a detached thread that calls `child.wait()` to reap it.
- In `svc_trigger` (`src/cron_jobs.rs`): same fix for the handler spawn path.

The fix is fire-and-forget — the reaper thread is detached and never blocks the request thread.

Closes #212

---
This pull request was opened by the **Implement my assigned issues without a PR (daemon + landing)** routine of moadim.


tupe12334:
Blocked: superseded by upstream. Zombie-reaping is already implemented via `spawn_and_reap`/`reap_in_background` in `src/utils/process.rs`, used by routine trigger. Needs a manual look (likely just closing).

— PR comment commander

PR #777, opened 2026-06-29
Reload routine/cron stores from disk on every GET

### LINKED ISSUE #774: GET endpoints serve stale startup cache; reload routine/job stores from disk on every read
## Problem

The daemon loads the routine store (`RoutineStore`) and cron-job store (`CronStore`) from disk **once at startup** into in-memory `Arc<Mutex<HashMap>>` caches (`routine_storage::load_store()` / `storage::load_store()`, wired in `build_app_with_shutdown` and `main`). Every read path serves from that cached snapshot:

- `routines::service::svc_list` / `svc_get` — back `GET /routines`, `GET /routines/{id}`, the iCal feed, and the MCP `list_routines` / `get_routine` tools.
- `cron_jobs::svc_list` / `svc_get` — back `GET /cron-jobs[/{id}]` and the MCP `list_cron_jobs` / `get_cron_job` tools.

So when the on-disk config changes underneath a running daemon, GET responses keep returning the **stale startup snapshot** until the daemon is restarted.

### Symptom

Editing a shared config repo and `git pull`-ing it into `~/.config/moadim` does not show up. Concretely: change a routine's `machines = [...]` targeting list, pull, and the running daemon still reports the old value on every GET. "I pulled the repo and don't see the changes."

## Expected

A GET reflects what is currently on disk. Pulling routine/job edits (including `machines`) into `~/.config/moadim` should be visible without restarting the daemon.

## Proposed fix

On every GET read (list + single-get, for both routines and cron-jobs, across HTTP, MCP, and iCal), re-scan the on-disk directory and refresh the in-memory store before serving. Disk is already the source of truth — every create/update/delete/trigger persists before returning — so replacing the map on read does not lose state.

Notes:
- Reload must go through the existing `load_*_from_dir` path so the gitignored `scheduled.local.toml` sidecar (`last_scheduled_trigger_at`) is preserved, not clobbered.
- Make the read path directory-aware (inject `routines_dir` / `jobs_dir` onto `AppState` and `MoadimMcp`) so the behavior is deterministic and unit tests run against temp dirs in parallel rather than the global `~/.config/moadim`.

## Acceptance

- A routine/job written to disk after startup appears in list/get without restart; one removed on disk disappears; a `machines` edit on disk is reflected on the next get.
- `last_scheduled_trigger_at` survives a reload.
- `cargo test` and `cargo clippy --all-targets -- -D warnings` stay green.

tupe12334:
## What

Reload the routine and cron-job stores from disk on **every GET** (list + single-get) across all surfaces — HTTP, MCP, and the iCal feed — so freshly pulled/edited config is reflected without a daemon restart.

Closes #774.

## Why

Both stores were loaded from disk **once at startup** into in-memory `Arc<Mutex<HashMap>>` caches; all reads (`svc_list`/`svc_get`) served that snapshot. Editing the shared config repo and `git pull`-ing it into `~/.config/moadim` had no effect on a running daemon — e.g. changing a routine's `machines = [...]` targeting list was invisible until restart.

## How

- **Reload.** `reload_store_from_dir(store, dir)` in `routine_storage.rs` and `storage.rs` re-scans `dir` and replaces the map in place under one `lock_recover()` (no partial store visible to a concurrent reader). The scan loop is shared with `load_store_from_dir`.
- **Base-dir-aware loaders.** `load_routine_from_base` / `load_job_from_base` resolve `routine.toml`, `state.local.toml`, `scheduled.local.toml` (and `job.toml`/`job.local.toml`) relative to a base dir, so any directory is self-contained — production or a test tempdir, no env mutation, parallel-safe.
- **Dir injection.** `AppState` and `MoadimMcp` carry `routines_dir` / `jobs_dir`, defaulting to `paths::routines_dir()` / `paths::jobs_dir()` in `build_app_with_shutdown`.
- **Wiring.** `routines::svc_list`/`svc_get`, `svc_ical`/`svc_ical_routine`, and `cron_jobs::svc_list`/`svc_get` take a `dir: &Path` and reload before serving. Routine `list`/`get`/`ical_feed` handlers switched to `State<AppState>`. MCP list/get tools pass the injected dirs. Create/update/delete/trigger unchanged (they persist before returning).
- Routine reload goes through the same load path, so the gitignored `scheduled.local.toml` (`last_scheduled_trigger_at`) is preserved on reload — covered by a test.

## Tests

`cargo test` → **727 passed, 0 failed**. `cargo clippy --all-targets -- -D warnings` clean (only the pre-existing `trunk not found` build-script note). Affected tests rewritten to write on-disk fixtures into a tempdir and pass that dir. New tests: routine/job written after store-build becomes visible; disk removal disappears; a `machines` edit is reflected on the next get; `last_scheduled_trigger_at` survives a reload.

## Tradeoffs / follow-ups

- **Per-GET disk re-scan.** Each read now does a `read_dir` + per-routine TOML parse (≈3 small files/routine) before serving — O(N files), unconditional. Fine at the expected scale (tens of routines/jobs); if a UI polls `GET /routines` heavily, a follow-up could add an mtime-based cache or debounce.
- **Disk is authoritative on read.** A routine in memory but absent on disk (mid-`git pull`, torn write) drops from the read until disk settles. `routine.toml` uses `atomic_write`; cron `job.toml` uses a plain `fs::write` (pre-existing).
- `migrate_routine_dirs_from_dir` still uses the global-path loader — intentional; it's a startup-only op on the real dir, not a read path.


tupe12334:
LGTM

Issue #723, opened 2026-06-26
Default "Update moadim cargo package" routine always generates with empty machines list


tupe12334:
## Problem

The default **\"Update moadim cargo package\"** routine is generated with an empty `machines` list. Since no machines are targeted, the routine never runs — defeating its purpose as a default.

## Expected Behavior

The default \"Update moadim cargo package\" routine should always include a default machines entry (e.g. match all / `*`) so it runs on every machine unless explicitly overridden by the user.

## Actual Behavior

Routine generates with `machines: []` — empty machines list. No machines targeted, routine never fires.

## Fix

Seed the default routine's `machines` field with a catch-all entry at generation time.

PR #781, opened 2026-06-30
fix(routines): repair empty machines list on legacy default routines (#723)

tupe12334:
## Problem

Default routines seeded before machine-awareness was introduced ended up with an empty `machines` list, leaving them permanently dormant — no machine ever matched them, so they never ran. This is issue #723.

The bug occurred because `reconcile` preserved the existing machines list verbatim (`machines: cur.machines.clone()`) and the `up_to_date` check did not consider an empty machines list as "drift", so the routine was silently skipped every startup and never repaired.

## Fix

Two minimal changes to `src/routines/defaults/mod.rs`:

1. **`up_to_date` check** now includes `&& !cur.machines.is_empty()`. An empty list is treated as drift so reconcile is triggered.
2. **Machine seeding in reconcile's output** — when `cur.machines.is_empty()`, seed the current machine (same as `materialize` does for fresh defaults) instead of cloning the empty list.

The repair runs automatically on the next daemon start without any user action.

## Tests added (`mod_tests.rs`)

- `reconcile_treats_empty_machines_as_drift_and_seeds_current_machine` — the core regression test
- `reconcile_returns_none_when_machines_already_set_and_otherwise_current` — ensures the fix doesn't over-trigger rewrites for healthy routines
- `materialize_assigns_non_empty_machines_list` — asserts the seeding invariant for fresh defaults

All 726 tests pass. `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` are both clean.

---

*This pull request was opened by the **Daemon + Landing auto-improve & auto-merge lint/docs PRs** routine of moadim. @tupe12334*


tupe12334:
LGTM

tupe12334:

PR #782, opened 2026-06-30
fix(defaults): re-seed machines when existing default has empty list

### LINKED ISSUE #723: Default "Update moadim cargo package" routine always generates with empty machines list

tupe12334:
Closes #723

## Root cause

When a default routine was first seeded before the `machines` field was introduced (PR #573), it had `machines: []`. On subsequent daemon restarts, `reconcile()` treated an empty `machines` list as a deliberate user choice and preserved it — so the routine continued to have no machines and never fired.

## Fix

Two changes to `reconcile`:

1. Added `&& !cur.machines.is_empty()` to the `up_to_date` guard — an existing default with an empty machines list is considered "needs rewrite" rather than "up to date."
2. On rewrite, if `cur.machines` is empty, seed it with `current_machine()` (matching what `materialize` does for a fresh default). If `cur.machines` is non-empty, carry it over unchanged as before.

This preserves user reassignments (e.g. moving the default to a different machine) while fixing the upgrade path for routines that predate the machines field.

## Tests

Two new unit tests in `mod_tests.rs`:
- `reconcile_re_seeds_empty_machines_with_current_machine` — the fix case
- `reconcile_preserves_non_empty_machines` — non-regression for user reassignments

725 tests pass.

---

*This pull request was opened by the Open issues → fix PRs (owned repos + my orgs) routine of moadim. @tupe12334*


tupe12334:
Blocked: superseded by upstream. `reconcile()` in `src/routines/defaults/mod.rs` (main) already re-seeds an empty `machines` list on a default routine — it treats an empty list as legacy drift and repairs it from the current spec, exactly this PR's fix. Needs a manual look before merge.

— PR comment commander

PR #791, opened 2026-06-30
fix(routines): surface agent_command_available distinct from agent_registered

### LINKED ISSUE #383: Routine health never checks the agent `command` binary exists on PATH — a routine with an uninstalled agent reports healthy yet silently no-ops
## Problem

A routine's agent health is reported solely by `agent_registered`, which is computed as:

```rust
// src/routines/model.rs
let agent_registered = agent_toml_path(&routine.agent).exists();
```

This only asserts that `~/.config/moadim/agents/<name>.toml` **exists**. It never verifies that the agent's actual `command` executable (e.g. `claude`, `codex`, `hermes`, or a custom one) is installed and resolvable on `PATH`.

So a routine whose agent config is present and well-formed, but whose `command` binary is **not installed**, will:

- pass create/update validation,
- report `agent_registered: true` (a healthy green dot in the UI),
- sync into the OS crontab and fire on schedule,
- launch `tmux new-session … <command> …` where `<command>` resolves to nothing — the session dies immediately with `command not found`.

The result is a **silent no-op**: the routine looks healthy everywhere the daemon reports status, but no agent ever runs. The only trace is a `command not found` line buried in the per-run `agent.log`.

This is the same failure class as #187 (tmux is an unchecked hard dependency), but for a **different** binary — the agent `command`, which `cron_path()` already resolves separately from tmux in `src/routines/command.rs`. It is also distinct from #301 (malformed TOML) and #139 (unknown agent *name*): here the config is valid and the agent name is known; only the underlying executable is missing.

## Motivation

Installing a routine for an agent CLI you haven't installed yet (e.g. creating a Codex routine on a box that only has `claude`) is a common, easy mistake. Today it fails invisibly until someone digs into `agent.log`. Surfacing "agent binary not found" at create time — and in the routine health signal — turns a silent dead routine into an actionable error, consistent with the project's recurring "don't show a healthy dot for something that will silently no-op" theme (#301).

## Acceptance criteria

- [ ] Resolve the agent's `command` against `PATH` (reuse the existing `bin_dir`/`cron_path` resolution logic in `src/routines/command.rs`) when building `RoutineResponse`.
- [ ] Expose the result distinctly from `agent_registered` — e.g. a new `agent_command_available: bool` field — so the UI/CLI can distinguish "config present" from "binary runnable". Keep `agent_registered` meaning "config file exists/parses" so existing semantics don't change.
- [ ] The UI routine row reflects an unhealthy/warning state when the agent binary is missing (no healthy green dot for a routine that will no-op).
- [ ] Optionally warn at routine create/update when the agent binary is absent (non-fatal — the binary may be installed later), surfaced in the API response and logs.
- [ ] If `moadim doctor` (#323) lands, it includes this check; the per-routine health signal here is independent of and complementary to that CLI command.
- [ ] Tests cover both states (binary present vs absent) using a `PATH`/probe seam, in line with the repo's test-seam conventions.

---

> _This issue was filed as a **suggestion** by the "Suggest new GitHub issues (daemon + landing)" routine of moadim. It is intentionally left unassigned — self-assign to approve and pick it up. cc @tupe12334_

tupe12334:
## Why

A routine's agent health today is reported solely by `agent_registered`, computed as `agent_toml_path(&routine.agent).exists()`. That only checks that `~/.config/moadim/agents/<agent>.toml` is present — it never verifies the agent's actual `command` executable (`claude`, `codex`, `hermes`, or a custom one) is installed and resolvable on `PATH`.

So a routine whose agent config is present and well-formed, but whose `command` binary isn't installed, will pass create/update validation, report `agent_registered: true` (a healthy green dot), sync into the crontab, and fire on schedule — only to launch `tmux new-session … <command> …`, where `<command>` resolves to nothing and the session dies immediately with "command not found." The result is a silent no-op: the routine looks healthy everywhere the daemon reports status, but no agent ever runs. The only trace today is a buried line in `agent.log`.

This is the same failure class as the existing `tmux_available` check (`src/routines/command.rs`), but for the agent binary instead of `tmux`.

## What

- Added `agent_command_available_in`/`agent_command_available` in `src/routines/command.rs`, generalizing the existing `tmux_available_in`/`tmux_available` pair (reusing the already-tested `bin_dir_in` resolution) to check an arbitrary command name against `PATH`.
- Added a new `agent_command_available: bool` field to `RoutineResponse` (`src/routines/model.rs`), computed in `from_routine` by loading the routine's agent config and resolving its `command` against `PATH`. Falls back to `false` when the agent config is missing, unreadable, or malformed (no `command` can be resolved in that case either) — kept distinct from `agent_registered`, whose semantics ("config file exists") are unchanged.
- This automatically threads through `GET`/`POST`/`PUT`/`DELETE` `/routines` (and the MCP equivalents, which serialize the same `RoutineResponse`), regenerating `apis/openapi.json` via the existing self-healing `committed_spec_is_current` test.
- Added unit tests in `command_tests.rs` (present/missing/live-PATH/unset-PATH branches, mirroring the existing `tmux_available*` tests) and `model_tests.rs` (registered+resolvable, registered-but-not-on-PATH — the key case this field exists to surface — and not-registered-at-all).

Deliberately scoped to the backend signal only (REST/MCP), not the UI/CLI surfacing called out as optional follow-up in the originating issue — keeping this PR small and reviewable.

## Verification (local)

```
cargo fmt --check                                                            # clean
cargo clippy --workspace --all-targets -- -D warnings                        # clean (root + ui crate)
cargo test                                                                    # 731 passed
cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs' --
  --skip restart::restart_tests::stop_running_and_wait_succeeds_without_pid_file_when_server_eventually_stops
                                                                               # 100% line coverage
```

(That one restart test is the pre-existing timing flake under coverage instrumentation already being fixed by open PR #783 — unrelated to this change, so it was skipped for the coverage run; plain `cargo test` passes it too most runs.)

Closes #383.

---

cc @tupe12334


tupe12334:
LGTM

tupe12334:

PR #792, opened 2026-06-30
fix(routines): populate handler registry from disk at startup

### LINKED ISSUE #790: Cron job handler_registered is always false — HandlerRegistry is never populated from disk

## Problem

`HandlerRegistry` (`Arc<HashSet<String>>`) backs `CronJobResponse.handler_registered` — `handlers.contains(&job.handler)` in `CronJobResponse::from_job` (`src/cron_jobs.rs:96-99`). The only constructor for that set is:

```rust
// src/cron_jobs.rs:161-163
pub fn new_registry() -> HandlerRegistry {
    Arc::new(HashSet::new())
}
```

This is exactly what production startup calls — `build_app_with_shutdown` (`src/routes/http.rs:258`, the only non-test call site, reached from the server's `main` startup path) always builds the `AppState`/MCP service with `new_registry()`, i.e. an **empty** set, and nothing ever inserts into it afterward.

`handlers_dir()` (`src/paths/mod.rs:73`) — the directory the registry is supposed to reflect — is never scanned to populate the registry. Its only production use is building a single handler's spawn path in `svc_trigger` (`src/cron_jobs.rs:378`) and in `src/sync/mod.rs:237,473` (unrelated to building the registry). Grepping every call site of `new_registry`/`handlers_dir` confirms: outside of test helpers (`cron_jobs_tests.rs`, `http_tests.rs`, `mcp_tests.rs`), the registry is never populated from disk.

**Result:** `GET /cron-jobs`, `GET /cron-jobs/{id}`, and the equivalent MCP tools report `handler_registered: false` for *every* job, in *every* real deployment, regardless of whether the handler script actually exists on disk.

This is a stronger claim than #16 item 2, which assumes the registry *is* scanned at boot and only asks to document that later additions aren't hot-reloaded — that assumption is wrong; there is no scan at all, at boot or otherwise, so the field is permanently useless as shipped today.

## Motivation

`handler_registered` exists so REST/MCP clients (and the agents the daemon serves) can validate a cron job's handler before relying on it. Right now the field is silent dead weight — it can never be `true` — which means:
- Clients cannot trust it to catch a genuinely missing/misnamed handler.
- It's actively misleading: a caller might reasonably read `false` as "handler missing" and chase a nonexistent bug instead of a registry bug.

## Acceptance criteria

- [ ] At daemon startup, scan `handlers_dir()` and populate `HandlerRegistry` with the discovered handler identifiers (consistent with the extension-stripping convention used elsewhere for handler resolution).
- [ ] `GET /cron-jobs` / MCP `list_cron_jobs`/`get_cron_job` return `handler_registered: true` for a job whose handler script genuinely exists on disk, verified by an integration test that builds production `AppState`/MCP state via the real startup path (not `new_registry()` directly — that's why this regressed unnoticed).
- [ ] Document the scan lifecycle (startup-only vs. hot-reload on later handler additions), closing the doc ambiguity raised in #16 item 2.

tupe12334:
## Why

`handler_registered` on `GET /cron-jobs` / `GET /cron-jobs/{id}` (and the equivalent MCP tool) is supposed to report whether a job's handler script exists in the server's handler registry, but it was **always `false`**: `build_app_with_shutdown` wired the app state's `HandlerRegistry` to the always-empty `new_registry()`, and nothing in the codebase ever scanned `~/.config/moadim/handlers/` to populate it outside of test helpers. This makes the field permanently misleading in every real deployment — fixes #790.

## What

- Added `populate_registry()` in `src/cron_jobs.rs`, which reads `paths::handlers_dir()` and collects the file names found there into a `HandlerRegistry`. Keyed by exact file name to match how `svc_trigger` resolves `job.handler` (`handlers_dir().join(&job.handler)`, no extension guessing). A missing/unreadable directory yields an empty registry rather than an error.
- Wired `populate_registry()` into `build_app_with_shutdown` (the single router-building function used by both production startup and tests) in place of `new_registry()`.
- Kept `new_registry()` (now `#[cfg(test)]`) since the rest of the test suite still uses it to construct an empty registry directly.
- Added two tests: registry is empty when the handlers directory doesn't exist, and registry correctly reports a handler as registered when its script file is present on disk.

## Testing

- `cargo build --all-targets`
- `cargo test` — 726 passed
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo fmt --check` — clean
- Added a `CHANGELOG.md` entry under `## [Unreleased] / ### Fixed` per repo convention (required by the `unreleased-entry` check since this touches `src/`).

---


tupe12334:
Blocked: superseded by upstream. `agent_registered` is already computed live per-response via `agent_toml_path(&routine.agent).exists()` in `src/routines/model.rs` — not a stale, never-populated registry, so the bug class this PR fixes doesn't exist in the current design. Needs a manual look (likely just closing).

— PR comment commander

Issue #16, opened 2026-06-11
Docs: behaviors that required source-diving to implement a working job


tupe12334:
While wiring up a cron job through the MCP + filesystem interface, several behaviors were undocumented (in README or tool descriptions) and only discoverable by reading the crate source. Documenting these would have made first-time implementation correct without trial-and-error.

### 1. Manual trigger does NOT execute the handler
`trigger_cron_job` (and `POST /cron-jobs/{id}/trigger`) only updates `last_triggered_at` — it does **not** run the handler script. `svc_trigger` just stamps the timestamp and persists. This is surprising: "trigger" strongly implies execution. There is currently no documented way to run a handler on demand for testing.
- **Fix:** state in the tool description that trigger only records the timestamp; document how (if at all) to force a real run. Consider an actual run-now option.

### 2. Handler registry is read only at server startup
Adding a script to `~/.config/moadim/handlers/` after the server is running leaves `handler_registered: false` until restart. Nothing documents that the registry (and the jobs dir) is scanned at boot, nor whether the file-watcher mentioned in the README ("watches for changes") covers handlers too.
- **Fix:** document the scan/watch lifecycle — what is hot-reloaded vs. requires restart.

### 3. `handler_registered: false` gives no remedy
The flag is returned but undocumented: what it means, why it would be false, and how to make it true (correct dir, no extension, executable bit, restart). 
- **Fix:** document the resolution rules and a checklist for when it is false.

### 4. Scheduled-execution semantics are undocumented
README says the server "watches for changes" and "execs it on each run", but there is no doc of: timezone used for cron evaluation, whether `seconds` precision actually fires, behavior when the machine is asleep/off at fire time, overlap handling if a previous run is still going, and where stdout/stderr land.
- **Fix:** a short "Execution model" section.

### 5. `job.log` is documented but not created until first run
README shows a `job.log` format, but the file does not exist until the handler has actually executed once. A note that it is created lazily would avoid confusion when checking for output.

### 6. No `list_handlers` over MCP
An agent creating a job has no way to discover valid handler identifiers — `list_cron_jobs` on an empty store returns `[]`, and the handlers dir is not exposed. (Cross-ref the schema gaps in #15.)

### 7. Cron format (seconds field) — also a tool-schema gap
Covered in #15, repeated here for completeness: the 6–7 field requirement is only in the README job.toml example, not in the `schedule` tool description.

[...]

PR #793, opened 2026-06-30
fix(cron): populate HandlerRegistry from disk at startup

### LINKED ISSUE #790: Cron job handler_registered is always false — HandlerRegistry is never populated from disk

tupe12334:
## What

`scan_registry()` reads `handlers_dir()` at startup and builds the `HandlerRegistry` from the (extension-stripped) filenames found there. `build_app_with_shutdown` now calls it instead of `new_registry()` (which just returns an empty `HashSet`).

## Why

`new_registry()` was the only thing production startup (`build_app_with_shutdown`, the only non-test call site) ever called to build `AppState`'s `HandlerRegistry`. Nothing scanned `handlers_dir()` to populate it. Result: `GET /cron-jobs`, `GET /cron-jobs/{id}`, and the equivalent MCP tools reported `handler_registered: false` for *every* job, in *every* real deployment, regardless of whether the handler script genuinely existed on disk — the field was permanently useless as shipped.

## Implementation

- `scan_registry()` reads `handlers_dir()`, keeps only files (`is_file()`, which follows symlinks — unlike `file_type()`, so a handler reached via a symlink is still discovered), and strips each entry's extension via `file_stem()` to match the extension-less `handler` field stored on a `CronJob` (mirroring the convention `sync::resolve_handler_path` uses to go the other way).
- `new_registry()` is now `#[cfg(test)]`-only — production no longer calls it, but it's still useful as an empty-registry baseline across the existing test suite.
- The registry stays read-only after startup (matching the existing `Architecture.md` claim and the `Arc<HashSet<...>>` type, which has no interior mutability) — a handler script added to the directory after the daemon starts is not picked up until the next restart. Documented in `Architecture.md` and the function's doc comment, closing the doc ambiguity raised in #16 item 2.

## Tests

- `cron_jobs::scan_registry_strips_extensions_and_skips_directories` / `scan_registry_returns_empty_set_when_handlers_dir_missing` — unit coverage for the scan logic.
- `routes::http::router_handler_registered_true_when_handler_script_exists_on_disk` — integration test that builds the production `AppState` via the real `build_app()` startup path (not `new_registry()` directly, which is exactly why this bug shipped unnoticed) and asserts `handler_registered: true` for a job whose handler script exists on disk.

## Checks

- `cargo fmt --check` — clean
- `cargo test` (single-threaded) — 729 passed
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — 100% line coverage
- `CHANGELOG.md` updated under `[Unreleased]`

Closes #790

---
This pull request was opened by the *Open issues → fix PRs (owned repos + my orgs)* routine of moadim.

### DIFF
--- a/Architecture.md
+++ b/Architecture.md
@@ -103,6 +103,8 @@ pub type HandlerRegistry = Arc<HashSet<String>>;
 
 Both are cloned into `AppState` (REST) and `MoadimMcp` (MCP). Every write acquires the mutex, updates in memory, then flushes to disk before releasing.
 
+`HandlerRegistry` is built once, at startup, by `scan_registry()` scanning `handlers_dir()` and stripping each entry's extension. It is **not** hot-reloaded: a handler script added to the directory after the daemon starts is not reflected until the next restart.
+
 ---
 
 ## Service layer
@@ -319,7 +321,7 @@ main()
   TcpListener::bind(:5784)
   routes::http::run_with_listener_until(store, listener, pending())
     build_app(store)
-      AppState { store, handlers: new_registry() }
+      AppState { store, handlers: scan_registry() }   scan ~/.config/moadim/handlers/ → HandlerRegistry
       StreamableHttpService::new(|| MoadimMcp::new(...))
       Router::new()  ← wire all routes + middleware
     banner::print(addr)          stdout: REST / MCP / UI URLs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,14 @@ Versions map to the `v*` git tags that drive the crates.io publish workflow.
 
 ### Fixed
 
+- `GET /cron-jobs`, `GET /cron-jobs/{id}`, and the equivalent MCP tools always
+  reported `handler_registered: false`, even when the job's handler script
+  genuinely existed on disk. The `HandlerRegistry` backing that field was built
+  with `new_registry()` at startup — an empty set that nothing ever populated
+  from `handlers_dir()`. The daemon now scans `handlers_dir()` at startup
+  (`scan_registry()`), so `handler_registered` reflects reality. The registry
+  stays read-only after startup; a handler script added later requires a
+  restart to be picked up. (#790)
 - `docs/moadim.1`'s `.TH` header reported a stale `moadim 0.16.0` even though
   `Cargo.toml` had moved on to 0.18.0 — the hand-maintained man page has no
   build-time link to the crate version, so a release could silently ship a man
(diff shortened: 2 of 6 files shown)


tupe12334:
- [ ]

Issue #504, opened 2026-06-21
Security: network-exposable daemon API/MCP has no authentication — setting MOADIM_BIND_ADDR to a non-loopback address exposes unauthenticated routine CRUD (= RCE)


tupe12334:
## Problem

The daemon serves a full read/write API (REST + MCP + the bundled UI) with **no authentication of any kind**. There is no `Authorization`/bearer/token/api-key check anywhere in the codebase — every endpoint is reachable by anyone who can open a TCP connection to the bound address.

That API is not read-only: it can **create, edit, trigger, and delete routines**, and a routine launches an arbitrary agent command in a workbench. In other words, write access to this API is **arbitrary code execution** on the host.

Today the only thing standing between that and the network is the default bind address:

```rust
// src/cli.rs
pub const BIND_ADDR: &str = "127.0.0.1:5784";
const BIND_ADDR_ENV: &str = "MOADIM_BIND_ADDR";

pub fn bind_addr() -> String {
    std::env::var(BIND_ADDR_ENV).unwrap_or_else(|_| BIND_ADDR.to_string())
}
```

`MOADIM_BIND_ADDR` is accepted verbatim with no validation and no warning. The moment an operator sets it to `0.0.0.0:5784` (to reach the UI from another machine, behind a reverse proxy, in a container, etc.) the entire routine-CRUD surface — i.e. unauthenticated RCE — is exposed to the LAN or the internet. The repo already treats the daemon as network-exposed (see `SECURITY.md`, the security-headers and path-leak hardening issues), so loopback-only is not a sufficient security boundary on its own.

## Motivation

- A single env var flips the daemon from "local-only" to "unauthenticated RCE for anyone who can route to it," with no guard rail and no log warning.
- Defense in depth: even on loopback, any local user / any process (incl. a browser visiting a malicious page that POSTs to `127.0.0.1:5784`, i.e. DNS-rebinding / CSRF-style attacks) can drive the API.
- All the existing hardening work (security headers #406, path-leak headers #356, world-readable perms #382) hardens a door that currently has no lock.

## Proposed approach

- Add an **optional shared-secret token** (e.g. `MOADIM_API_TOKEN`, or a generated token persisted under the config dir) enforced via a small auth middleware (sibling to `logger` / `fs_location`), checking `Authorization: Bearer <token>` (and/or `X-Moadim-Token`).
[...]

Issue #406, opened 2026-06-19
HTTP responses carry no security headers — add X-Frame-Options / nosniff / Referrer-Policy / CSP to the served UI & API (clickjacking + MIME-sniff hardening)


tupe12334:
## Problem
The daemon serves a full browser dashboard (the ~1.1 MB SPA at `/`, plus the client-routed pages `/cron-jobs`, `/routines`) and an unauthenticated REST API on the same loopback port. HTTP responses currently carry **no security response headers** — a grep over `src/` finds no `X-Frame-Options`, `Content-Security-Policy`, `X-Content-Type-Options`, or `Referrer-Policy` anywhere. The only response-header middleware (`src/middlewares/fs_location.rs`) *adds* `x-server-root` / `x-server-exe-dir` (the headers #356 wants removed); none harden the response.

Because the dashboard's buttons drive an unauthenticated loopback API (create / trigger / delete routines, `POST /shutdown`), the missing headers leave two cheap, real vectors open:

- **Clickjacking** — a malicious page can `<iframe src="http://localhost:5784/">` the dashboard and trick the operator into clicking *Trigger* / *Delete* / *Stop*. With no `X-Frame-Options` / CSP `frame-ancestors`, framing is allowed.
- **MIME sniffing** — without `X-Content-Type-Options: nosniff`, a browser may content-sniff a response into an unintended type.

This is complementary to (not a duplicate of) #266 (Host/Origin validation against DNS rebinding) and #356 (removing the path-leaking headers): those address *who can reach* the API and *what leaks*; this adds standard *defense-in-depth response headers* the daemon emits today on nothing.

## Motivation
- Defense-in-depth for a daemon that intentionally ships a browser UI bound to loopback, where the same port also exposes state-changing, unauthenticated endpoints.
- Cheap, well-understood hardening (a single tower/axum response-header layer) that mirrors the security posture already pursued elsewhere in the repo (#266, #356, #382) and on the landing repo.
- Closes the clickjacking gap on the dashboard's destructive controls (Trigger / Delete / Shutdown).

## Acceptance criteria
- A response-header layer applies to all HTTP responses (UI + `/api/v1`), setting at minimum:
  - `X-Frame-Options: DENY` (and/or CSP `frame-ancestors 'none'`)
  - `X-Content-Type-Options: nosniff`
  - `Referrer-Policy: no-referrer` (or `strict-origin-when-cross-origin`)
[...]

PR #800, opened 2026-06-30
fix(security): harden CSP beyond frame-ancestors (#551)

### LINKED ISSUE #551: Security: CSP is `frame-ancestors`-only — add script-src/style-src/default-src to harden the unauthenticated dashboard against in-page injection (in-code follow-up to #406)
## Problem

`src/middlewares/security_headers.rs` sends exactly one CSP directive on every response:

```
content-security-policy: frame-ancestors 'none'
```

The code comment is explicit that this is deliberately partial:

> The CSP is intentionally scoped to `frame-ancestors` only [...] A fuller `script-src`/`style-src` policy is left as follow-up once it can be verified against the bundled UI.

This is the follow-up. There is currently **no** `default-src`, `script-src`, `style-src`, `object-src`, `base-uri`, or `form-action`, so an injected inline `<script>` or a `<base>` tag is not blocked at all.

That matters here more than for a typical app: the served dashboard drives an **unauthenticated loopback API with destructive controls** — create / trigger / delete routines and `POST /shutdown`. Any in-page script execution (a reflected/stored XSS, or a compromised front-end dependency) runs with full authority to schedule and trigger agent sessions that carry the operator's tokens — the same severity class as the auth gaps (#504, #253) and the Host/Origin / DNS-rebinding gap (#266). A real `script-src` closes the in-page-injection → destructive-action path even if those are fixed.

## Motivation

Defense-in-depth that complements, not replaces, the auth work: `frame-ancestors` already blocks clickjacking (#406); a `script-src`/`object-src`/`base-uri` policy blocks the remaining in-page execution and URL-hijack vectors.

## Why it is non-trivial (must be verified against the bundled UI)

- `prebuilt.html` ships an **inline `<script type="module">`** that instantiates embedded base64 WASM. Plain `script-src 'self'` is insufficient — it needs `'wasm-unsafe-eval'` plus either a `sha256-…` hash of the inline bootstrap or a per-response nonce.
- The UI loads a **Google webfont** (`fonts.googleapis.com` / `fonts.gstatic.com`). `style-src` / `font-src` must allow those **until #467 self-hosts the font** — coordinate with #467 (ideally self-host first, then `font-src 'self'`).
- **Swagger UI** (`utoipa-swagger-ui`) has its own inline styles/scripts and must keep working under the chosen policy.

## Acceptance criteria

- [ ] CSP extended beyond `frame-ancestors` with, at minimum: `default-src 'none'`; a working `script-src` for the SPA bootstrap (hash or nonce) + `'wasm-unsafe-eval'`; `style-src`; `font-src`; `img-src 'self' data:`; `connect-src 'self'`; `base-uri 'none'`; `form-action 'none'`; `object-src 'none'`; `frame-ancestors 'none'` retained.
- [ ] Bundled Yew/WASM SPA loads and functions with **no CSP violations** in the browser console.
- [ ] Swagger UI loads and "Try it out" works with no CSP violations.
- [ ] Webfont handled: allowed in `style-src`/`font-src`, or (preferred) self-hosted via #467 with `font-src 'self'`.
- [ ] A middleware test asserts the full CSP header value on a served response (extend `security_headers_tests.rs`).
- [ ] Any directive that must stay loose (e.g. `'unsafe-inline'` in `style-src`) is documented inline with its reason, so it is an explicit decision rather than an oversight.

---

_This issue was filed as a suggestion by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. It is left unassigned — self-assign to approve it. cc @tupe12334_

tupe12334:
## Summary

The dashboard's `Content-Security-Policy` header previously carried only `frame-ancestors 'none'` (the anti-clickjacking fix from #406), leaving `script-src` / `style-src` / `default-src` completely unset. Since the served dashboard drives an **unauthenticated loopback API with destructive controls** (create / trigger / delete routines, `POST /shutdown`), an injected inline `<script>` or a `<base>` tag was not blocked at all — this PR closes that gap.

## What changed

`src/middlewares/security_headers.rs` now sends:

```
default-src 'self';
script-src 'self' 'unsafe-inline' 'wasm-unsafe-eval';
style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;
font-src 'self' https://fonts.gstatic.com;
img-src 'self' data:;
connect-src 'self';
base-uri 'none';
form-action 'none';
object-src 'none';
frame-ancestors 'none'
```

This was verified against the actual served UI (built `cargo build`/`cargo test` locally and inspected `prebuilt.html` plus the vendored Swagger UI dist):

- **`script-src`/`style-src` keep `'unsafe-inline'`**: the bundled Yew/WASM SPA (`prebuilt.html`, produced by `src/build/ui.rs`) self-inlines its entire wasm-bindgen JS glue + base64 WASM payload into one `<script type="module">` so the daemon ships as a single self-contained file. That payload's bytes change on every UI rebuild, so a `sha256-…` CSP hash would need to be computed and wired through the build script rather than hardcoded — left as further follow-up rather than expanding this PR's scope. The page also has one inline `<style>` block, and the embedded Swagger UI's React components set inline `style="…"` attributes at runtime.
- **`script-src` also carries `'wasm-unsafe-eval'`**: required for the SPA's `WebAssembly.instantiate` call. This is narrower than `'unsafe-eval'` — it permits WASM compilation without permitting `eval()`/`Function()`. (Confirmed Swagger UI's bundled JS only uses `new Function(...)` inside a try/catch'd `globalThis` polyfill that already falls back gracefully, so it doesn't need `'unsafe-eval'`.)
- **`style-src`/`font-src` still allow `https://fonts.googleapis.com`/`https://fonts.gstatic.com`**: contrary to the issue's premise, the webfont-inlining fix (#467) has **not** landed on `main` yet — it's still open as PR #519 — so the dashboard genuinely still fetches its webfont from Google Fonts today. These CDN allowances should be dropped once #519 merges.
- **`img-src` allows `data:`**: the dashboard CSS uses an inline SVG `data:` URI for a noise-texture background.
- Inspected the actual vendored Swagger UI dist (`utoipa-swagger-ui` v9.0.2 / swagger-ui v5.17.14, downloaded at build time): its `index.html` has zero inline scripts/styles — `swagger-ui-bundle.js`, `swagger-ui-standalone-preset.js`, and the generated `swagger-initializer.js` are all served as separate same-origin files, compatible with `script-src 'self'`.
- `frame-ancestors 'none'` is unchanged.

Updated `security_headers_tests.rs` and `http_tests.rs` to assert the full new CSP header value (previously asserting the old `frame-ancestors 'none'`-only string).

## Why

Closes #551. Defense-in-depth complementing (not replacing) the loopback-API auth work tracked separately (#504, #253, #266): `frame-ancestors` already blocks clickjacking; this closes the remaining in-page-script-execution and URL-hijack vectors so an XSS/compromised dependency can't silently drive the destructive routine/shutdown controls.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings` (the exact command `.github/workflows/lint.yml`'s `clippy` job runs)
- [x] `cargo test` — all 726 tests pass, including the updated CSP assertions
- [x] Verified no unrelated files changed (in particular, `prebuilt.html`, which a local `trunk`-equipped `cargo build` regenerates as a build artifact, was reverted before committing)

---

cc @tupe12334


tupe12334:
LGTM

tupe12334:

PR #802, opened 2026-07-01
fix(routines): derive workbench launch path from paths::workbenches_dir()

tupe12334:
## Why

The generated cron launch command (`build_routine_command` in `src/routines/command.rs`) hardcoded:

```sh
```

instead of resolving the base through `paths::workbenches_dir()` — the single seam that `moadim_home()` / `paths::home()` already funnel through, and which **honors `MOADIM_HOME_OVERRIDE`** (added in #585). Every *consumer* of workbenches already goes through this seam:

- `routines/cleanup/mod.rs` — the TTL reaper sweep
- `routines/service.rs` — listing runs for the LOGS view and reading `agent.log`

But the *producer* — the shell command that actually creates and writes into the workbench on each cron fire — used a hand-written `$HOME/.moadim/workbenches` literal instead.

The two only coincided because, today, `paths::home()` falls back to `dirs::home_dir()` (≈ `$HOME`). The moment `MOADIM_HOME_OVERRIDE` is set on the daemon process (or the layout changes in the future), they diverge silently:

- Runs get written under the override path
- ...but the reaper and LOGS view keep scanning `$HOME/.moadim/workbenches`
- Result: workbenches leak forever (invisible to the reaper) and the LOGS view shows nothing for real runs — a classic two-sources-of-truth drift with no compile-time or test guard.

Filed as GitHub issue #601 (auto-suggested); this PR implements the fix.

## What changed

- `src/routines/command.rs`: `build_routine_command` now resolves `workbenches_base` via `crate::paths::workbenches_dir()` and shell-quotes it into the `WB=` assignment, instead of the hardcoded `$HOME/.moadim/workbenches` literal. `$SLUG-$TS` is still appended at run time, so the default (no override) install behaves identically.
- `src/routines/command_tests.rs`: added `build_routine_command_workbench_base_tracks_moadim_home_override`, which sets `MOADIM_HOME_OVERRIDE` to a tempdir and asserts the generated command's `WB=` base matches `paths::workbenches_dir()` (and that the old hardcoded literal is gone), guarding against this regressing again.
- `CHANGELOG.md`: added a `### Fixed` entry under `[Unreleased]`.

## Verification

- `cargo fmt --check` — clean
- `cargo test` — 727 passed (one pre-existing, unrelated flaky timing test in `restart::restart_tests` intermittently fails locally — already tracked by open PR #783, "widen restart-test timing margins to stop flaking under cargo llvm-cov"; unaffected by this change and reproduces on `main` as well)

## Scope note

This does not duplicate any of the ~30 currently open PRs from this routine (clippy lint enablement, CI coverage gate, UI features, cron/routine handler-registry population, CSP hardening, PATCH-alias docs, zombie reaping, rmcp bump, etc.) — this is a distinct correctness bug in the `MOADIM_HOME_OVERRIDE` path-resolution seam.

---


tupe12334:
LGTM

tupe12334:

tupe12334:

PR #817, opened 2026-07-01
fix(routines): make slugify Unicode-aware so non-Latin titles keep their content

### LINKED ISSUE #262: slugify drops all non-ASCII characters — non-Latin routine titles (Hebrew/CJK/Cyrillic) collapse to "routine", blocking a second such routine
## Problem

`slugify` (`src/routines/command.rs:12`) keeps only characters for which `char::is_ascii_alphanumeric()` is true:

```rust
for ch in title.chars() {
    if ch.is_ascii_alphanumeric() {
        out.push(ch.to_ascii_lowercase());
        ...
```

Every non-ASCII letter — Hebrew, Cyrillic, Greek, CJK, and even Latin letters with diacritics (`é`, `ü`, `ñ`) — is treated as a separator and dropped. A title made **entirely** of such characters slugifies to the empty string and falls through to the `"routine"` fallback.

Concrete cases:

- `"עדכון יומי"` → `"routine"`
- `"Café Report"` → `"caf-report"` (the `é` is silently lost)
- `"日次レポート"` → `"routine"`
- `"Отчёт"` → `"routine"`

The slug is not cosmetic — it names the on-disk workbench dir (`$HOME/.moadim/workbenches/$SLUG-$TS`) and the tmux session (`moadim-$SLUG-$TS`) in `build_routine_command`, and the routine folder via `routine_dir(&slug)`.

Two real consequences:

1. **A second non-Latin-titled routine cannot be created at all.** `svc_create` (`src/routines/service.rs:108`) rejects a create whose slug matches an existing routine's slug:
   ```
   a routine with the name "routine" already exists
   ```
   So once any all-non-ASCII-titled routine exists, every subsequent one (with a distinct, perfectly valid title) is refused with a 409 — the slug namespace for non-Latin titles has exactly one slot (`routine`).
2. **Unreadable, undebuggable names.** The one routine that does get created lands in `~/.moadim/workbenches/routine-<ts>` with tmux session `moadim-routine-<ts>`, giving no hint which routine it belongs to when inspecting `tmux ls` or the workbenches dir.

This is distinct from existing issues: #226 is about *empty/whitespace* titles, and #188 is about *ASCII* slugs that collide and silently overwrite. Here the title is non-empty and valid, but the ASCII-only slugifier erases it. The fix is a Unicode-aware slugifier (`#188`'s collision disambiguation alone would still leave every non-Latin routine named `routine-<uuid>`).

## Motivation

- The repository owner works in Hebrew (routine prompts and titles are routinely non-Latin), so this is hit on the very first Hebrew-titled routine and blocks the second one.
- Internationalized titles are a normal, valid input — silently mangling or rejecting them is surprising and hard to diagnose (the API/UI list keys on UUID and looks fine while disk/tmux state is degraded).

## Acceptance criteria

- [ ] `slugify` preserves meaningful content from non-ASCII titles instead of dropping it. Pick one approach:
  - make it Unicode-aware (keep Unicode alphanumerics via `char::is_alphanumeric`, normalizing/lowercasing), or
  - transliterate to ASCII (e.g. a `deunicode`/`any_ascii`-style mapping) before slugifying.
- [ ] Two routines with distinct non-Latin titles produce distinct, non-`"routine"` slugs and can both be created (no spurious 409).
- [ ] Latin-with-diacritics titles (`"Café"`, `"Über"`) no longer silently lose characters.
- [ ] The `"routine"` fallback still applies only when *nothing* usable remains (e.g. a title of only punctuation/whitespace) — preserving #226's separate concern.
- [ ] Tests cover Hebrew, CJK, Cyrillic, and accented-Latin titles for both slug content and create-uniqueness; existing ASCII slug behavior is unchanged.
- [ ] Folder-naming/slug rules in `Architecture.md` note the Unicode handling.

---
*This issue was filed as a suggestion by the "Suggest new GitHub issues (daemon + landing)" routine of moadim. It is left unassigned — self-assign to approve. cc @tupe12334*

tupe12334:
## Why

`slugify` (`src/routines/command.rs`) only kept characters passing `char::is_ascii_alphanumeric()`. Every non-ASCII letter — Hebrew, Cyrillic, Greek, CJK, and even accented Latin (`é`, `ü`) — was treated as a separator and dropped:

- `"עדכון יומי"` → `"routine"`
- `"日次レポート"` → `"routine"`
- `"Отчёт"` → `"отчёт"` → previously `"routine"`
- `"Café Report"` → previously `"caf-report"` (the `é` silently lost)

The slug isn't cosmetic: it names the on-disk workbench dir (`$HOME/.moadim/workbenches/$SLUG-$TS`) and the tmux session (`moadim-$SLUG-$TS`), and `svc_create` rejects a create whose slug collides with an existing routine's. Once one all-non-Latin-titled routine exists, every subsequent one — with a distinct, perfectly valid title — was refused with a 409, because the ASCII-only slugifier gave every one of them the same `"routine"` slug.

## What changed

`slugify` now uses `char::is_alphanumeric()` / `char::to_lowercase()` (Unicode-aware) instead of the ASCII-only variants, so non-Latin titles and accented Latin keep their meaningful content instead of collapsing to the generic fallback. The `"routine"` fallback still applies only when nothing usable remains (pure punctuation/whitespace titles), matching existing behavior for that case.

The slug is shell-quoted everywhere it's embedded into a crontab/shell command (`shell_quote(&slug)` in `build_routine_command`), so non-ASCII bytes there were already safe — this only changes what content is preserved, not how it's escaped.

Added tests for Hebrew, CJK, and Cyrillic titles, plus a check that two distinct non-Latin titles now produce distinct (non-`"routine"`) slugs, so a second such routine can actually be created.

Fixes #262.

## Verification

- `cargo fmt --check` — clean
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — passes (100% line coverage); confirmed the one intermittent failure I hit (`restart::restart_tests::stop_running_and_wait_succeeds_without_pid_file_when_server_eventually_stops`) is a known flake unrelated to this change, already tracked by open PR #783
- Pushed via the repo's own pre-push hook (fmt/clippy/coverage), which passed

---


tupe12334:
LGTM

tupe12334:

tupe12334:

Issue #399, opened 2026-06-19
HTTP responses are never compressed — the ~1.1 MB single-page UI ships uncompressed on every load/refresh (add a tower-http CompressionLayer)


tupe12334:
## Problem

The web UI is a single self-contained `index.html` (the Yew SPA with its JS shim and base64-inlined wasm bundle, built in `src/build/ui.rs` and served via `include_str!` from `GET /` in `src/routes/http.rs`). The committed artifact is ~1.1 MB (`prebuilt.html` is 1,158,519 bytes), and the served `OUT_DIR/index.html` is the same order of magnitude.

The Axum router in `src/routes/http.rs` applies **no HTTP response compression** — `tower-http` is not even a dependency (only `tower` is, as a dev-dependency). So every UI load and every SPA-route refresh ships the full ~1 MB uncompressed over the wire. HTML + JS + base64 wasm are highly compressible (typically 70–85% smaller with gzip, more with brotli), so this is a large, repeated, easily-avoided transfer cost — meaningful when the daemon is reached over a LAN / reverse proxy / slow link, and even on loopback it inflates initial paint.

## Motivation

- Cut the SPA payload ~3–5× on every load/refresh with a one-layer change.
- Cheap, standard, low-risk: a `tower-http` `CompressionLayer` negotiates via `Accept-Encoding` and is a no-op for clients that don't advertise it.
- Benefits the same UI surfaces recently invested in (logs view, calendar, modals).

## Proposed approach

- Add the `tower-http` dependency with the `compression-gzip` (and optionally `compression-br`) feature.
- Add `CompressionLayer::new()` to the router in `src/routes/http.rs` (outermost relevant layer, so the large `/` HTML body and the OpenAPI JSON under `/docs` are compressed; the Swagger UI assets benefit too).
- Keep it content-negotiated so non-gzip clients (and existing tests) still get identity encoding.

## Acceptance criteria

- [ ] `GET /` with `Accept-Encoding: gzip` returns `Content-Encoding: gzip` and a body materially smaller than the identity response.
- [ ] `GET /` without `Accept-Encoding` still returns the full uncompressed HTML (no regression for plain clients).
- [ ] A test asserts the compressed response is present/smaller (e.g. compare body length with vs. without the header).
- [ ] `tower-http` is pinned consistently with the repo's dependency conventions; `fmt`/`clippy`/`missing_docs` gates stay green.

---
[...]

PR #834, opened 2026-07-01
feat(http): add ETag + 304 revalidation for the embedded SPA

### LINKED ISSUE #401: UI served with no Cache-Control/ETag — the ~1.1 MB SPA is re-downloaded in full on every load/refresh (add an ETag + 304 revalidation)
## Problem

The web UI is served by `GET /` (and the SPA fallback) from `index()` in `src/routes/http.rs`, which returns the embedded `index.html` (the ~1.1 MB single-page bundle with inlined wasm/JS) as a plain `Html<&'static str>`. No `Cache-Control`, `ETag`, or `Last-Modified` header is attached anywhere — `grep -rn 'Cache-Control\|ETag\|If-None-Match\|304' src` returns nothing, and the router only stacks the `fs_location` and `logger` middleware layers.

Result: every page load, refresh, and client-side navigation that hits the server re-downloads the full ~1.1 MB body. The browser has no validator to revalidate against, so even when the bundle is byte-identical it is transferred again in full.

## Motivation

This is the other half of #399. That issue reduces *transfer size* per load via compression; this issue avoids the transfer *entirely* on repeat loads. The two are complementary:

- The embedded asset is a compile-time constant (`include_str!(concat!(env!("OUT_DIR"), "/index.html"))`), so it has a trivially strong validator: a hash of the bytes computed once at startup.
- With an `ETag`, a refresh sends `If-None-Match` and the daemon answers `304 Not Modified` with an empty body — saving the full payload on every repeat visit, on top of whatever compression buys.
- For a local-loopback daemon UI that operators keep open and refresh frequently, this is a meaningful latency and bandwidth win.

## Acceptance criteria

- [ ] `GET /` (and the SPA-fallback HTML responses) emit a strong `ETag` derived from the embedded `index.html` bytes (computed once, e.g. a hash of the constant), plus a `Cache-Control` header.
- [ ] A request carrying a matching `If-None-Match` receives `304 Not Modified` with an empty body; a non-matching / absent one receives `200` with the full body and the `ETag`.
- [ ] Caching headers do **not** break the JSON API or the long-lived `/mcp` SSE stream (scope the behavior to the static UI response, not blanket-applied).
- [ ] Tests cover both the `200 + ETag` and the `304` revalidation paths.
- [ ] Plays well with #399: compression still applies to the `200` body.

---

_This issue was filed as a suggestion by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. It is intentionally left unassigned — approve it by self-assigning. cc @tupe12334_

tupe12334:
## Summary

- `GET /` (and the SPA fallback used for client-routed paths like `/routines`, `/cron-jobs`) served the full ~1.1 MB embedded `index.html` on every single load and refresh, with no caching signal — the gzip `CompressionLayer` (#399) shrinks the transfer but the browser still re-fetches unconditionally every time.
- Adds a strong `ETag` computed once from the embedded HTML's content (deterministic across restarts of the same binary; changes only when a new build embeds different bytes) and honors a matching `If-None-Match` request header with a bodyless `304 Not Modified`.
- Sends `Cache-Control: no-cache` so the browser always revalidates (rather than trusting a local TTL, since the content can change on any daemon upgrade) — revalidation is cheap (a small request/response), while the actual 304 avoids re-transferring the multi-hundred-KB body when nothing changed.
- Closes #401.

## Why

Every dashboard load or SPA-route refresh currently re-downloads the entire embedded UI bundle for no reason — the content is static per binary build. This is a small, self-contained fix with no behavior change to the app itself, just standard HTTP caching semantics for a static asset.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --workspace` (full suite, including 3 new tests: ETag header present + `Cache-Control: no-cache`, matching `If-None-Match` → bodyless `304`, stale `If-None-Match` → normal `200`)
- [x] `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — 100% line coverage maintained
- [x] Added a `CHANGELOG.md` entry under `## [Unreleased] / ### Added`

cc @tupe12334


tupe12334:
LGTM

tupe12334:

tupe12334:
This PR is approved (LGTM). I merged main into the branch to refresh a stale CodeQL result and fixed a compile break the merge exposed (a stale 2-arg `build_app()` test call from before main's signature refactor — pushed as a6d5d7f). Remaining blocker: `unreleased-entry` fails — this branch predates main's migration to Changesets (`pnpm changeset`) and has no `.changeset/*.md` file, which the new gate requires.

Issue #175, opened 2026-06-17
Tests mutate the real user crontab (missing CRONTAB_BIN/PATH isolation)


tupe12334:
## Problem

Several unit tests in `src/routines/service_tests.rs` invoke the **real** system `crontab` binary, clobbering the developer's live crontab.

`svc_create` / `svc_update` / `svc_delete` call `sync_routines_to_crontab` → `write_crontab` → `crontab -`. The binary is resolved via `crontab_bin()` (`src/sync/mod.rs:142`):

```rust
fn crontab_bin() -> String {
    std::env::var("MOADIM_CRONTAB_BIN").unwrap_or_else(|_| "crontab".to_string())
}
```

Tests are meant to isolate this two ways:
- set `MOADIM_CRONTAB_BIN` to a fake shim (`CronShim` / `FailingCrontabShim`), or
- clear `PATH` via `with_empty_path(...)` so `crontab` can't be spawned and sync fails harmlessly.

These tests do **neither**, so they hit `/usr/bin/crontab`:

| Test | service_tests.rs | Sync trigger |
|------|------------------|--------------|
| `svc_update_sets_ttl_secs` | 150 | `svc_update` (succeeds → syncs) |
| `svc_create_rejects_duplicate_slug` | 73 | first `svc_create` + `svc_delete` |

(`svc_update_rejects_renaming_into_existing_slug:112` short-circuits on `Conflict` before sync, so it is currently safe but should be wrapped defensively.)

## Impact

Running the test suite locally **overwrites the developer's real user crontab**. Observed in the wild: a developer's crontab routines block was replaced with a single stale line from the `svc_update_sets_ttl_secs` fixture:

```
@daily /bin/sh -l '/Users/<u>/.config/moadim/routines/svc-update-ttl-zzz/run.sh' # moadim-routine:ttl-id
```

`ttl-id` / slug `svc-update-ttl-zzz` come directly from `make_routine("ttl-id", "Svc Update Ttl ZZZ", ...)`. The test removes the routine dir afterward but **not** the crontab line, leaving a dangling block and dropping every real routine until the daemon next re-syncs.

## Fix

Wrap the offending tests' `svc_*` calls in `with_empty_path(|| { ... })`, matching the sibling `*_warns_when_crontab_sync_fails` tests (service_tests.rs:320/347/375/403).

PR #869, opened 2026-07-02
fix(routines): structurally guard the sh spawn seam against test builds

### LINKED ISSUE #217: Routine trigger and cron handler spawn real processes (sh / handler) with no structural test seam
## Problem

Two spawn sites run real external processes, isolated in tests only by convention (empty `PATH` / fake paths), not structurally:

- `src/routines/service.rs:211` — `svc_trigger` runs the agent command via `sh`. Tests cover the failure branch by clearing `PATH` (`svc_trigger_warns_when_spawn_fails`).
- `src/cron_jobs.rs:334` — spawns the cron `handler_path`. Tests pass non-existent handler paths (the `cov-handler ... No such file` output seen in test runs).

A future test that triggers a routine/handler without clearing `PATH` or using a fake path could execute a real agent command or handler script on the developer's machine.

## Severity

LOW — no persistent system state mutated; current tests isolate via empty `PATH` / fake paths.

## Fix (optional / defense-in-depth)

Centralize external-process spawning through a single helper that, under `cfg(test)`, refuses to spawn real binaries unless an explicit per-call override is set — mirroring the `crontab_bin()` guard (#211). This would cover `sh`, the handler, tmux, and kill in one place. Lower priority than the launchctl/plist siblings.

Related: #175, #211.

tupe12334:
## Why

`spawn_routine_command` (`src/routines/service.rs`) launches the routine command via `Command::new("sh")` with no structural guard — tests isolate it only by convention (clearing `PATH`). A future test that triggers a routine without clearing `PATH` could execute a real login shell, and thus a real agent launch, on the developer's machine. `crontab_bin()` (`src/sync/mod.rs`) already closes this same class of gap for crontab I/O (issue #175); this seam had no equivalent.

## What

- Added `sh_bin()` in `src/routines/service.rs`, mirroring `crontab_bin()`: honors a `MOADIM_SH_BIN` env override, and in test builds defaults to a nonexistent path when unset so a spawn fails harmlessly regardless of `PATH` state.
- `spawn_routine_command` now spawns via `sh_bin()` instead of the bare `"sh"` literal.
- Added `sh_bin_never_resolves_to_real_sh_in_test_builds` (structural guard) and `sh_bin_honors_override` (override branch) unit tests.

Out of scope: the issue also mentions a second historical spawn site (`cron_jobs.rs`) and `tmux`/`kill`. `cron_jobs.rs` no longer exists in this repo (routines are now scheduled by invoking the `moadim` binary directly, not a generated per-routine script); `tmux` is invoked from inside the generated shell command, not via a direct `Command::new` in Rust; and `kill` already has its own override (`MOADIM_KILL_BIN`, `src/restart.rs`) that intentionally does *not* default-deny in test builds, since a real kill test exercises it against its own spawned child. The `sh` seam was the one remaining unguarded site.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` (643 passed, 100% line coverage)
- [x] Changeset added (`.changeset/harden-sh-spawn-seam.md`)

Closes #217

---

This pull request was opened by the Implement my assigned issues without a PR (daemon + landing) routine of moadim.

@tupe12334 for visibility.


tupe12334:
LGTM

tupe12334:

PR #873, opened 2026-07-02
fix(routines): reject and normalize blank machines entries

### LINKED ISSUE #600: machines targeting list is not validated — empty/whitespace entries silently exclude a routine/job (and an all-empty list evades the dormant warning)

tupe12334:
## Summary

`machine::targets` matches a routine's `machines` targeting list by **exact string equality** against the resolved machine name, but the list is stored verbatim on create/update — no trim, no empty-filter, no dedup — unlike every sibling field (`title`, `repositories`, `tags`). Two silent-failure modes follow:

1. A whitespace-padded or typo'd entry (`" laptop"`, trailing `\n`) never equals the resolved machine name, so the routine is silently excluded from that machine's crontab.
2. An entry that's just `""` makes the list *non-empty*, so it slips past the dormant-routine warning (which only fires on `machines.is_empty()`) — the routine then runs **nowhere, with no warning at all**, which is strictly worse than the empty-list case.

This adds `validate_machines` (mirroring the existing `validate_repositories` / `validate_tags` helpers) and wires it into both `svc_create` and `svc_update` in `src/routines/service.rs`. It rejects blank/whitespace-only entries with an indexed `400`, and trims + dedupes accepted entries. Both REST (`src/routines/handlers.rs`) and MCP (`src/routes/mcp.rs`) route through this shared service layer, so both surfaces are covered by one change.

Closes #600.

## Test plan
- [x] `cargo build`
- [x] `cargo test` (651 passed, including 3 new tests: `svc_create_rejects_blank_machine`, `svc_create_trims_and_dedupes_machines`, `svc_update_rejects_and_sets_machines`)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (clean)
- [x] `cargo fmt --check` (clean)
- [x] `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` (100% line coverage, passes)
- [x] Added a changeset (`.changeset/validate-routine-machines.md`, patch)


tupe12334:
LGTM

tupe12334:

Issue #462, opened 2026-06-20
feat(routines): opt-in missed-run catch-up — cron silently skips a routine's fire time when the host sleeps/daemon is down; run it on next startup


tupe12334:
## Problem

moadim schedules every routine by writing a line into the OS crontab. Plain `cron`/`launchd` calendar timers have no catch-up semantics: if the host is **asleep, powered off, or the daemon is down** at a routine's scheduled fire time, that occurrence is **silently skipped** — it never runs late, and there is no record that it was missed.

For a tool whose pitch is *"Cron jobs that run while you sleep"* / *"Set the loop. Forget the keyboard."*, this is a real reliability hole on the exact hardware most users run it on (laptops/desktops that sleep). A daily 06:00 routine simply does not run on any day the machine happened to be closed at 06:00, and the operator gets no signal that it was skipped.

This is distinct from the existing backlog:
- #394 re-syncs the crontab **block** on startup (ensures future lines exist) — it does **not** execute a fire time that already elapsed while down.
- #290 (`ignore_until` snooze) and #435 (global pause) intentionally *suppress* runs — the opposite concern.
- #453 records a run's exit status, but a **missed** run produces no run at all, so there is nothing to record.

## Motivation

Unattended, scheduled agent work is only trustworthy if a missed window is either recovered or at least surfaced. Anacron solves exactly this for periodic cron jobs on machines that aren't always on; moadim has no equivalent. Without it, "runs while you sleep" quietly degrades to "runs only if the machine happened to be awake."

## Proposal

Add an **opt-in, per-routine** missed-run catch-up. On daemon startup (and after resume from sleep, where detectable), for each enabled routine compute whether its most recent scheduled fire time falls between `last_scheduled_trigger_at` (see #155) and now; if so, and catch-up is enabled, launch a single make-up run.

Sketch:
- New optional routine field, e.g. `catch_up: bool` (default `false`, so current behavior is unchanged) — or an enum (`off` / `last_missed` / `all_missed`) if firing every skipped occurrence is ever wanted; default to collapsing to **at most one** make-up run to avoid a thundering herd after a long downtime.
- Persist `last_scheduled_trigger_at` (depends on / pairs with #155) so "what was the last fire we actually ran" is known across restarts.
- On startup, for each enabled routine with `catch_up` on: if `next_fire_after(last_scheduled_trigger_at) <= now`, enqueue one make-up run, then resume normal scheduling.
[...]

PR #874, opened 2026-07-02
fix(routines): don't stamp last_scheduled_trigger_at on a manual trigger

### LINKED ISSUE #478: Manual trigger pollutes last_scheduled_trigger_at: svc_trigger runs the shared command, which writes the scheduled-fire sidecar — conflating manual and scheduled runs (#128/#155)

tupe12334:
## Problem

`svc_trigger` (the manual trigger path — `POST /routines/{id}/trigger`, MCP `trigger_routine`) spawns the exact same shell command the crontab runs, via `build_routine_command`. That shared command builder unconditionally emitted a statement writing `last_scheduled_trigger_at` into the routine's `scheduled.local.toml` sidecar, which the daemon reads straight back on load (`read_scheduled_state` in `src/routine_storage.rs`). So a manual, on-demand trigger falsely recorded that the cron schedule had fired, conflating manual and scheduled runs (undoing the distinction established by #128/#127/#155) and risking misleading missed-run/catch-up logic (#462) and observability.

## Fix

Threaded a `stamp_scheduled: bool` flag through `build_routine_command` (`src/routines/command.rs`) and the shared `spawn_routine_command` helper (`src/routines/service.rs`):

- `svc_trigger` (manual) now calls `spawn_routine_command(&routine, false)` — the generated shell script omits the `scheduled.local.toml` write entirely.
- `svc_trigger_scheduled` (the daemon-side endpoint the crontab line drives via `moadim schedule trigger`) calls `spawn_routine_command(&routine, true)` — unchanged behavior, the sidecar stamp is still written exactly as before.

This is a root-cause fix in the one shared command builder rather than a patch on the manual-trigger caller alone, so no other caller of `build_routine_command` can reintroduce the bug.

## Testing

- Added `build_routine_command_omits_scheduled_sidecar_stamp_for_manual_trigger` (`src/routines/command_tests.rs`): asserts the generated command for `stamp_scheduled: false` contains no reference to `last_scheduled_trigger_at` or the sidecar path.
- Added `svc_trigger_manual_leaves_scheduled_sidecar_untouched_but_scheduled_path_writes_it` (`src/routines/mod_tests.rs`): an end-to-end regression test that actually spawns both a manual (`svc_trigger`) and scheduled (`svc_trigger_scheduled`) trigger with a harmless `true` agent command, waits for the shells to finish, and asserts the `scheduled.local.toml` sidecar is absent after the manual trigger and present after the scheduled one — directly covering the acceptance criteria.
- Updated all other `build_routine_command` call sites (existing tests) for the new parameter.

Verified locally:
- `cargo fmt --check` — clean
- `cargo test` — 667 passed, 0 failed
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — 100% line coverage, gate passes
- `typos` — clean

Closes #478

---

This pull request was opened by the Open issues → fix PRs (owned repos + my orgs) routine of moadim.


tupe12334:
Blocked: superseded by upstream. Same as #541/#490 — `svc_trigger`/`svc_trigger_scheduled` (main) are already cleanly separated so a manual trigger never stamps the scheduled sidecar. Needs a manual look before merge.

— PR comment commander

Issue #742, opened 2026-06-26
feat(ui): model field on routine create/edit — free-text input alongside agent selector


tupe12334:
## Problem

The routine create/edit form lets users pick an **agent** but has no way to specify which **model** the agent should use. Users are locked into whatever model the agent config defaults to, with no per-routine override.

## Proposed solution

Add a **Model** field to the routine create/edit form — a free-text input (not a dropdown) so any model ID can be typed in (e.g. `claude-sonnet-4-6`, `claude-opus-4-8`, custom/local model names, etc.).

### Data layer

- Add `model: Option<String>` to `Routine`, `CreateRoutineRequest`, and `UpdateRoutineRequest` in `src/routines/model.rs`.
- Pass the model value through to the agent invocation in the command builder (`src/routines/command.rs`).

### UI layer (`ui/src/routines.rs`)

- Add a **Model** text input field in the create/edit modal, below the Agent selector.
- Field is optional — blank means "use agent default".
- Display the model value in the routine detail/row view alongside the agent name.

## UX notes

- Free-text (not a select) because model IDs change frequently and vary by provider.
- Placeholder text: e.g. `claude-sonnet-4-6` (or leave blank to use agent default).
- No validation beyond non-empty-when-present; the agent itself will surface unknown model errors at runtime.

tupe12334:
Closing as already implemented on main: routines have `model: Option<String>` through persisted/API request/response types, agent command invocation appends `--model <id>`, and the React create/edit form includes an optional MODEL field below AGENT.

PR #833, opened 2026-07-01
feat(ui): add TTL preset row (1h/1d/7d/30d) to the routine form

tupe12334:
## Why

The `WORKBENCH TTL` field in the routine form takes a raw second count typed from memory (e.g. `604800` for 7 days). The `SCHEDULE` field right above it already solves the same usability problem with one-click cron presets — TTL had no equivalent, and it was a standing item in `TODO.md`.

## What

- Add a `1h` / `1d` / `7d` / `30d` preset row under the `WORKBENCH TTL` input, reusing the existing `.preset-btn` styling and mirroring `set_preset`'s pattern for the cron presets.
- New `.ttl-presets` CSS rule in `ui/index.html` (identical layout to `.cron-presets`).
- Removed the corresponding line from `TODO.md` per the file's own convention, and added a `CHANGELOG.md` entry under `## [Unreleased] / Added`.
- Regenerated `prebuilt.html` via `cargo build` (trunk installed locally) so the committed fallback UI reflects this change — it also picks up other UI work already merged to `main` that the committed bundle had fallen behind on.

## Verified locally

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test` (726 passed), `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` (gate passes), and `typos` all pass.

---
This pull request was opened by the Daemon + Landing auto-improve & auto-merge lint/docs PRs routine of moadim.

PR #988, opened 2026-07-05
feat(ui): add Model field to routine create/edit form

tupe12334:
## Why

`Routine`, `CreateRoutineRequest`, and `UpdateRoutineRequest` already carry an optional `model: Option<String>` override on the backend, and `src/routines/command.rs` already passes it through as `--model <id>` on the agent invocation — this landed already. But the UI form (`ui/src/routines/form.rs`) and table row (`ui/src/routines/row.rs`) never picked it up, so there is currently no way for a user to actually set a routine's model from the dashboard. This closes that gap (#742).

## What changed

- `ui/src/routines/model.rs`: add `model: Option<String>` to the UI's `Routine`, `CreateRoutineRequest`, and `UpdateRoutineRequest` mirrors of the server API.
- `ui/src/routines/form.rs`: add a free-text **MODEL** input below the Agent selector (optional; blank = agent default), following the same state/callback shape as the existing `goal` field.
- `ui/src/routines/page.rs`: forward the model value on edit-save, always sending it (like `goal`) so clearing the field clears it server-side too.
- `ui/src/routines/row.rs`: show the model under the agent name in the Routines table when set.
- Updated the `Routine`-literal test helpers in `filter_tests.rs`, `filter_health_tests.rs`, `state_tests.rs`, `command_palette_tests.rs`, and `schedule_heatmap_tests.rs` for the new field.
- Regenerated `prebuilt.html` via `trunk build` so the embedded UI bundle matches.
- Added a changeset.

## Test plan

- [x] `cargo test` (workspace, 875 passed) and `cargo test -p ui` (255 passed)
- [x] `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails, but not on this PR's own diff — same cross-platform WASM build reproducibility gap as #833 (introduced by #809). Verified: a local rebuild of this exact commit (`cargo check`, trunk, macOS) reproduces `prebuilt.html` byte-for-byte identically to what's committed — the bundle is functionally fresh, just not byte-identical to whatever CI's Ubuntu runner produces. All other checks are green. Needs a manual look (either re-running prebuilt-ui.yml on this exact commit to compare, or pinning trunk's version / investigating wasm-bindgen determinism) before merge.

— PR comment commander

PR #999, opened 2026-07-05
fix(cli): rotate daemon.log instead of letting it grow forever

### LINKED ISSUE #316: daemon.log grows without bound — no rotation/size cap, and every /health poll logs 2 info lines (idle UI ≈ 5.7k lines/day)
## Problem

The backgrounded daemon redirects **both** stdout and stderr into `~/.config/moadim/daemon.log`, opened in pure append mode and **never rotated or size-capped**:

```rust
// src/cli.rs  (spawn_detached)
let out = std::fs::OpenOptions::new()
    .create(true)
    .append(true)        // append forever — never truncated, never rotated
    .open(&log_path)?;
let err = out.try_clone()?;
```

Every detached start/restart appends to the same file, and nothing ever trims it. Over a long-lived install this file grows without bound until it fills the disk — there is no max size, no `.1`/`.2` roll, and no retention window.

The growth is amplified by the request logger, which logs **every** HTTP request at `info`:

```rust
// src/middlewares/logger.rs
log::info!("{} {}", method, path);          // 1 line in
log::info!("  -> {} {} in {}ms", ...);      // 1 line out
```

The web UI polls `GET /api/v1/health` every 30 s (`ui/src/main.rs`: `TimeoutFuture::new(30_000)`), and `moadim status --wait` polls in a loop too. A single open UI tab therefore writes ~2 lines / 30 s ≈ **5,760 lines/day** of pure health-check noise straight into `daemon.log` — drowning genuinely interesting events (routine triggers, errors, crontab-sync warnings) and accelerating the unbounded growth.

This is the daemon's *own* log file and is distinct from the per-run agent capture tracked in #268 (`agent.log` via tmux `pipe-pane`).

## Motivation

- A daemon meant to run unattended for weeks/months must not silently grow a log file until the disk is full.
- Operators reading `daemon.log` (the path `moadim` prints as `logs  …`) need signal, not thousands of `/health` lines per day.
- Mirrors the bounded-logging direction already taken for agent output (#268/#279/#280).

## Acceptance criteria

- [ ] `daemon.log` is bounded: either a size cap with rotation (e.g. roll to `daemon.log.1` at N MB, keep K files) or a documented truncate-on-start policy — pick one and document it in `README`/`Architecture.md`.
- [ ] Health-check / polling requests (`GET /health`) no longer log two `info` lines per hit — log them at `debug`/`trace`, sample them, or skip them in the request logger, so an idle-but-open UI doesn't flood the log.
- [ ] Non-health requests and mutations still log at `info` (no regression in useful coverage).
- [ ] Tests cover the rotation/cap boundary (rolls at the threshold; old segment pruned) and assert the request logger downgrades/skips `/health`.

---
*This issue was opened by the "Suggest new GitHub issues (daemon + landing)" routine of moadim, as a non-binding suggestion. It is intentionally left unassigned — self-assign it to approve. cc @tupe12334*

tupe12334:
## Summary

The backgrounded daemon redirects stdout/stderr into `~/.config/moadim/daemon.log`, opened in pure append mode with no size cap or rotation. Over a long-lived install this file grows without bound until it fills the disk.

(The other half of this issue — the request logger writing two `info` lines per `GET /health` poll — is already fixed; `src/middlewares/logger.rs` downgrades that path to `debug`.)

## Change

`spawn_detached_with` (used by both `spawn_detached` and `spawn_restart`) now checks the log file's size before opening it for append, and rotates it to a `daemon.log.1` sibling (replacing any previous one) when it exceeds 10 MiB. This is a rotate-on-(re)start policy: simple, matches the existing best-effort style of this module, and needs no background timer.

## Testing

- `cargo test --bin moadim cli_system_tests::` — 4 new tests (missing file is a no-op, small file left alone, oversized file rolled to `.1`, a stale `.1` is replaced) — all pass.
- `cargo clippy --bin moadim --tests` and `cargo fmt --check` — clean.

Closes #316


tupe12334:
LGTM

tupe12334:
Blocked: despite the LGTM, this can't be mechanically merged. `main` already ships an identical fix — `rotate_daemon_log_if_oversized`/`MAX_DAEMON_LOG_BYTES` in `src/cli_system.rs`, called from `spawn_detached_with` before opening `daemon.log`, byte-for-byte the same logic this PR adds. Merging as-is causes a duplicate-definition compile error (`E0428`), which is why `cargo test`/`clippy`/`cargo doc` are failing on this PR's own diff. Nothing left for this PR to contribute. Needs a manual look (likely just closing).

— PR comment commander

tupe12334:

PR #1002, opened 2026-07-05
feat(ui): expose Snoozed, Flagged, Agent-unregistered in STATUS filter dropdown

tupe12334:
## Summary

- Three `RoutineStatusFacet` variants (`Snoozed`, `HasFlags`, `AgentUnregistered`) were fully implemented in the filter logic but had no corresponding option in the STATUS dropdown, making them unreachable from the UI
- Adds "Snoozed", "Flagged", and "Agent unregistered" options to the STATUS `<select>` in the filter bar

## Test plan

- [ ] Existing filter roundtrip tests cover the codecs for all three variants
- [ ] Select "Snoozed" in the STATUS dropdown — only snoozed routines appear
- [ ] Select "Flagged" — only routines with open flags appear
- [ ] Select "Agent unregistered" — only routines whose agent config is missing appear
- [ ] "CLEAR" button resets back to "All"


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails, but not on this PR's own diff — same cross-platform WASM build reproducibility gap as #988/#833 (introduced by #809). Needs a manual look (either re-running prebuilt-ui.yml on this exact commit to compare, or pinning trunk's version / investigating wasm-bindgen determinism) before merge.

— PR comment commander

PR #1015, opened 2026-07-05
fix(build): regenerate stale prebuilt.html, document regen in CONTRIBUTING

### LINKED ISSUE #557: CI never verifies the committed prebuilt.html matches ui/ source — a stale UI bundle silently ships to every `cargo install` user
## Problem

`build.rs` → `src/build/ui.rs` embeds the daemon's web UI. When `trunk` is present it runs `trunk build --release`, inlines the JS/WASM into a single self-contained HTML, and then **copies that result back into the source tree at `prebuilt.html`** (`std::fs::copy(&output, &prebuilt)`). When `trunk` is *absent* — notably the `cargo install moadim` path and any contributor without trunk — the build silently falls back to the **committed** `prebuilt.html` (1.15 MB).

Nothing verifies that the committed `prebuilt.html` matches what `ui/` currently produces:

- The `Lint` and `Test` workflows deliberately do **not** install trunk or rebuild the UI — their comments state they rely on the committed `prebuilt.html` fallback to type-check/test the server crate.
- So a PR that changes `ui/src/**`, `ui/index.html`, or `ui/Trunk.toml` but does **not** regenerate `prebuilt.html` (e.g. authored on a machine without trunk) goes green while shipping a stale bundle.
- Result: every `cargo install moadim` user can receive a UI that lags behind the actual `ui/` source, with **zero CI signal** that the embedded dashboard is out of date.

This is distinct from existing issues:
- **#412** (CI never lints/compiles the `ui` crate) is about clippy/compile of the Yew crate — a UI that *compiles fine* can still ship a stale `prebuilt.html`, so that gate would not catch this.
- **#425** (`.gitattributes` for `prebuilt.html`) is about diff noise / linguist stats, not freshness.
- **#329** (prebuilt **binaries** on Releases) is about the install-without-Rust path, not the embedded UI artifact.

## Motivation

`prebuilt.html` is the exact artifact `cargo install` users get. A drift between it and `ui/` is invisible until someone manually rebuilds with trunk and notices a diff — i.e. it surfaces in production, not on the PR that caused it. A small CI guard turns a silent, user-facing staleness bug into a red check on the offending PR.

## Acceptance criteria

- A CI job (or a step in an existing workflow) that, on PRs touching `ui/**` and/or `prebuilt.html`:
  - installs `trunk` + the `wasm32-unknown-unknown` target,
  - runs the same UI build `build.rs` performs (so `prebuilt.html` is regenerated),
  - fails if the working tree's `prebuilt.html` differs from the committed copy (e.g. `git diff --exit-code prebuilt.html`).
- The failure message tells the contributor how to regenerate (`cargo install trunk && cargo build`, then commit `prebuilt.html`).
- Deterministic output: if `trunk`/wasm-bindgen versions make the inlined bytes non-reproducible across environments, document the pinned versions used (or compare a normalized form) so the check doesn't flap.
- CONTRIBUTING.md notes that `prebuilt.html` is a generated, committed artifact that must be regenerated when `ui/` changes.

---
*This issue was filed as a SUGGESTION by the "Suggest new GitHub issues (daemon + landing)" routine of moadim. It is intentionally left unassigned — self-assign it to approve. @tupe12334*

tupe12334:
## Context

Issue #557 asked for CI that verifies the committed `prebuilt.html` matches
what `ui/` currently produces. Investigating before writing anything new
turned up that **PR #809 already added this exact check**
(`.github/workflows/prebuilt-ui.yml`: installs trunk + wasm32 target, runs
`cargo check` so `build.rs` regenerates `prebuilt.html`, fails on
`git diff --exit-code -- prebuilt.html`, path-filtered to `ui/**` /
`src/build/ui.rs` / `prebuilt.html`) — merged ~14 hours before this PR. So
most of the acceptance criteria were already satisfied on `main` and this PR
does not re-add that job.

Two things were still wrong, found by actually exercising the check locally
with `trunk` installed:

1. **`main`'s committed `prebuilt.html` is stale right now**, despite the
   guard existing. `prebuilt-html-fresh` is currently *failing* on merged
   PRs #988, #1002, #1003, #1013 (all touch `ui/`) — but `main` has no branch
   protection (`GET /branches/main/protection` → 404), so the failing check
   doesn't block merges and they went in anyway. I confirmed this is real
   drift, not local-toolchain noise: two independent `trunk build --release`
   runs on my machine produced byte-identical output, and both differ from
   the currently committed file. This PR regenerates `prebuilt.html` from
   current `ui/` source to fix the immediate drift.
2. **CONTRIBUTING.md never got the acceptance-criteria note** that
   `prebuilt.html` is a generated, committed artifact that must be
   regenerated when `ui/` changes. Added next to the existing
   `apis/openapi.json` convention bullet.

## Not in scope of this PR (flagging, not fixing)

The root reason drift keeps recurring isn't a missing CI job anymore — it's
that `prebuilt-html-fresh` (and every other check) is advisory, not required,
because `main` has no branch protection rule. That's a repo-wide governance
change affecting every workflow, not a "straightforward CI diff-check" fix,
so I left it alone rather than guess at the right required-checks policy.
Recommend a follow-up: add branch protection on `main` requiring at least
`prebuilt-html-fresh`, `rustfmt`, `clippy`, and `cargo test` to pass before
merge.

## Testing

- `cargo check --workspace` — clean
- `cargo fmt --check` — clean
- `actionlint` — clean (existing `prebuilt-ui.yml` workflow, unchanged)
- Ran the workflow's actual logic locally end-to-end (had `trunk` available):
  installed the `wasm32-unknown-unknown` target, ran `cargo build` (triggers
  `build.rs` → `trunk build --release` → inline → copy to `prebuilt.html`),
  diffed against the committed copy — reproduced the same "drifted" failure
  GitHub Actions is currently reporting on #988/#1013, confirming the CI
  logic and my fix are both correct. Did not re-test the "matches" case via
  an actual Actions run (would require pushing and waiting on CI), but the
  local repro plus the two prior real CI runs on this same workflow are
  strong evidence it behaves correctly in both directions.
- Did not attempt any branch-protection change — flagged above instead.

Closes #557


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails, but not on this PR's own diff — it's the same cross-environment WASM build reproducibility gap already flagged on #833: `prebuilt-ui.yml` runs `cargo install --locked trunk` against the ubuntu-latest `stable` toolchain at CI run time, which resolves to different trunk/wasm-bindgen versions than whatever environment produced this PR's committed `prebuilt.html` — so the byte-for-byte comparison is inherently unstable across time/platform, not a sign this PR is actually stale. All other checks are green, and the CONTRIBUTING.md documentation change is sound. Needs a manual look (pinning trunk's exact version in the workflow, or making the freshness check tolerant of reproducible-but-not-identical output) before this or any `ui/`-touching PR can merge cleanly.

— PR comment commander

Issue #477, opened 2026-06-20
UI: show a humanized retention countdown ("expires in 2d" / "expired") per finished run in the routine LOGS view


tupe12334:
> Suggestion filed by the **Suggest new GitHub issues (daemon + landing)** routine of moadim — left unassigned for the maintainer to approve by self-assigning.

## Problem
The routine **LOGS** view lists a routine's finished runs and their captured output, but it gives no indication of *when each run's workbench will be reaped*. Each finished run's `agent.log` / `prompt.md` live in `~/.moadim/workbenches/<slug>-<ts>` and are deleted once the routine's effective TTL (`ttl_secs`) elapses, yet the UI surfaces nothing about that deadline. An operator opening LOGS can't tell whether a run's output is about to vanish or has already been reaped — they only discover it's gone after a cleanup sweep (cf. #357, which refreshes stale output but still shows no expiry).

This is an explicit item on `TODO.md`:
> Show a humanized retention countdown ("expires in 2d" / "expired") per finished run in the routine LOGS view, derived from the run's finish time and the routine's effective TTL.

## Motivation
- **Predictability** — operators know how long a run's logs remain available before deciding whether to copy them out.
- **Trust** — pairs naturally with #357 (stale-output refresh): instead of silently disappearing, a run visibly counts down to "expired".
- **No backend cost** — the value is derived client-side from data the UI already has (run timestamp + the routine's `ttl_secs`); no new endpoint required.

## Scope / notes
- Distinct from #327 (TTL *input* preset row), #311 (freed-bytes on cleanup), and #357 (refresh stale view) — this is a per-run **read-only countdown label**.
- The anchor for the countdown should be the run's *finish* time; note the open caveat in #174 (retention is currently measured from run *start*, not finish). The badge should reflect whatever effective-TTL anchor the daemon actually uses so the displayed countdown matches real reap behavior — coordinate with #174 if that anchor changes.
- Falls back gracefully when `ttl_secs` is unset (routine uses the default TTL) or when the run is already past its deadline (show "expired").

## Acceptance criteria
- [ ] Each finished run row in the routine LOGS view shows a humanized retention label derived from the run's effective expiry time (e.g. "expires in 2d", "expires in 47m", "expired").
- [ ] The label uses the routine's effective TTL (explicit `ttl_secs` or the documented default) and the same time anchor the cleanup watchdog uses, so the countdown matches actual reap behavior.
[...]

PR #1016, opened 2026-07-05
feat(routines): humanized retention countdown per finished run in HISTORY

tupe12334:
## Why

TODO.md and issue #477 both ask for the same thing: "Show a humanized retention countdown ('expires in 2d' / 'expired') per finished run ... derived from the run's finish time and the routine's effective TTL." Today the run-history (HISTORY) view lists each run's status/duration/exit code, but the only place retention is surfaced at all is the routine-level TTL number on the Routines table — a user has to manually add that to a run's finish time to know when its log actually disappears.

## What changed

- `RunSummary` (server) gains `retention_expires_at: Option<u64>` — `finished_at + Routine::effective_ttl_secs()` for a run whose workbench is still on disk, `None` while running or once the workbench is already gone (a run restored from `runs.log` after TTL reaping — nothing left to count down to there).
- `svc_list_runs` (single-routine history) populates it from the owning routine's TTL; the fleet-wide `svc_list_all_runs` passes `None` since the overview's "recent runs" panel is out of scope for this change.
- UI: `RunSummary` mirrors the new field, and the HISTORY page (`ui/src/routines/history.rs`) renders a new `RETENTION` column via a small pure formatter (`fmt_retention`), e.g. "expires in 12m" / "expires in 2h 5m" / "expired".

## Testing

- Added `fmt_retention` unit tests (under-a-minute, minutes, hours, at-deadline, past-deadline) in `ui/src/routines/history_tests.rs`, mirroring the existing `fmt_run_duration` test style.
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (912 root + 260 ui tests), and `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main.rs'` (100.00% lines) all pass.
- One pre-existing, unrelated finding: `src/routines/service_trigger_tests.rs` is already at 742 lines on `main` (over the local `linecheck --max-lines 700` pre-push gate), introduced by a prior commit and untouched by this diff (confirmed via `git diff main...HEAD -- src/routines/service_trigger_tests.rs` — empty). This gate isn't enforced in CI, only the local hook, so it doesn't block this PR, but is worth a follow-up split like #1008 did for three other files.
- Regenerated `apis/openapi.json` (self-healing test) and `prebuilt.html` (via `trunk`, matching how prior UI-touching PRs like #988 committed the rebuilt bundle).
- Added a changeset (`.changeset/run-history-retention-countdown.md`).

---

cc @tupe12334


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails, but not on this PR's own diff — same cross-environment WASM build reproducibility gap already flagged on #833/#1015: `prebuilt-ui.yml`'s unpinned `cargo install --locked trunk` against ubuntu-latest `stable` produces a different trunk/wasm-bindgen output than whatever environment generated this PR's committed `prebuilt.html`, so the byte-for-byte diff is unstable regardless of the actual UI change. All other checks (including the new `history_tests.rs` coverage) are green. Needs a manual look (pin trunk's version in the workflow, or relax the freshness check) before merge.

— PR comment commander

PR #1036, opened 2026-07-06
test(ui): cover parse_cron/describe_cron_live in cron_utils.rs

tupe12334:
## Summary

`ui/src/cron_utils.rs` holds the cron-expression parsing/formatting logic shared by the routines, schedule, and heatmap views (`parse_cron`, `describe_cron_live`, `reltime`). It had **zero** tests, unlike its two sibling pure-logic modules — `schedule.rs` (`schedule_tests.rs`) and `schedule_heatmap.rs` (`schedule_heatmap_tests.rs`) — which both follow a documented host-tested convention.

This PR adds `ui/src/cron_utils_tests.rs`, testing:
- `parse_cron`: empty/blank input, 5-field passthrough, 6-field-with-seconds (parses as-is via croner, preserving second-level detail), 7-field (seconds+year stripped to match the server's `normalize_schedule` in `src/utils/cron.rs`), `@keyword` schedules, and invalid input.
- `describe_cron_live`: the blank-input placeholder, the invalid-expression message, and the valid-expression description.

`reltime` is deliberately left untested — it calls `js_sys::Date::now()` and needs a wasm/DOM host to run, mirroring the pure/DOM split already documented in `refresh.rs` (`fmt_freshness` is host-tested, the DOM-touching parts aren't).

No behavior change — regression tests only. `ui` isn't held to the 100%-line-coverage gate (that's root-package-only per `CONTRIBUTING.md`), so this test gap was easy to miss; the diff is small and self-contained.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo test --workspace` (268 passed, up from 259; 9 new tests added, 0 failed)
- [x] `linecheck --max-lines 500` over `src/` and `ui/src/`
- [x] Added `.changeset/cron-utils-host-tests.md` (patch, package `moadim`)

### DIFF
--- a/.changeset/cron-utils-host-tests.md
+++ b/.changeset/cron-utils-host-tests.md
@@ -1,0 +1,17 @@
+---
+"moadim": patch
+---
+
+test(ui): cover `parse_cron`/`describe_cron_live` in `cron_utils.rs`
+
+`cron_utils.rs`'s field-count normalization (5-field passthrough, 6-field
+with seconds, 7-field seconds+year stripping, `@keyword`, and invalid input)
+and `describe_cron_live`'s validity/description pairing had no tests at
+all, unlike the sibling `schedule.rs`/`schedule_heatmap.rs` pure-logic
+modules which both have dedicated `*_tests.rs` files. Added
+`cron_utils_tests.rs` following that same host-tested convention.
+`reltime` is left untested — it calls `js_sys::Date::now()` and needs a
+wasm/DOM host, mirroring the pure/DOM split already documented in
+`refresh.rs`.
+
+No behavior change — regression tests only.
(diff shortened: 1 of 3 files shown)


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails, but not on this PR's own diff — same cross-environment WASM build reproducibility gap already flagged on #833/#1015/#1016: `prebuilt-ui.yml`'s unpinned `cargo install --locked trunk` against ubuntu-latest `stable` produces a different trunk/wasm-bindgen output than whatever environment produced the committed `prebuilt.html`, so the byte-for-byte diff is unstable regardless of this PR's actual (test-only) change. All other checks are green. Needs a manual look (pin trunk's version in the workflow, or relax the freshness check) before merge.

— PR comment commander

PR #1039, opened 2026-07-06
fix(security): warn at startup when binding to a non-loopback address

tupe12334:
## Why

The REST/MCP API has no authentication (#504): setting `MOADIM_BIND_ADDR` to a non-loopback address silently exposes unauthenticated routine CRUD to anyone who can reach that address. Full auth is a much larger change tracked separately in #504. Until that lands, this risk should at least be visible.

## What

- Adds `cli::bind_addr_is_loopback(addr: &str) -> bool`, a small helper that parses the configured bind address and checks `IpAddr::is_loopback` (treats an unparsable address as non-loopback, so callers warn rather than stay silent).
- `run_server()` in `main.rs` now logs a `log::warn!` at startup when the bind address isn't loopback-only, matching the existing tmux/python3 startup-check pattern already in that function.
- Two unit tests for the new helper (IPv4/IPv6 loopback, non-loopback, and unparsable input).

## Not in scope

This is a visibility fix, not authentication. Implementing real auth for the REST/MCP surface is tracked in #504 and is intentionally left out of this small PR.

## Testing

- `cargo fmt --check` — clean
- `cargo test --workspace` — 924 passed (was 922; +2 new tests), 0 failed
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — 100% line coverage (the new warning branch lives in `main.rs`, which is excluded from the floor; the new testable helper in `cli.rs` is fully covered)
- `typos` — clean on changed files

### DIFF
--- a/.changeset/warn-non-loopback-bind.md
+++ b/.changeset/warn-non-loopback-bind.md
@@ -1,0 +1,7 @@
+---
+"moadim": patch
+---
+
+### Fixed
+
+Warn at startup when the server binds to a non-loopback address. The REST/MCP API has no authentication (#504), so exposing it beyond `127.0.0.1`/`::1` grants anyone who can reach that address unauthenticated routine CRUD; the daemon now logs a loud warning at launch, matching the existing tmux/python3 startup checks, instead of leaving this risk silent.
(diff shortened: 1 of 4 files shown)


tupe12334:
Blocked: `linecheck` check is failing — `src/cli.rs` is 508 lines (max 500), caused by this PR's added helper + tests. Needs a source split/refactor, which is a judgment call beyond automated fix scope here.

PR #1068, opened 2026-07-07
chore(lint): enable clippy::redundant_closure_for_method_calls in ui crate

tupe12334:
## Why

The `ui` (Yew/WASM) crate has its own `[lints.clippy]` table (`all = "deny"` only) — it doesn't inherit the workspace root's extended deny-list via `[lints] workspace = true`. So every lint the repo's ongoing "enable clippy::X" series has added to root `Cargo.toml` (this one, `redundant_closure_for_method_calls`, since #549) has silently never applied to `ui/src`, even though CI's `clippy` job runs `--workspace` and does compile/lint `ui`.

## What

- Add `redundant_closure_for_method_calls = "deny"` to `ui/Cargo.toml`.
- Fix the 3 violations it surfaces (`ui/src/routines/form.rs` x2, `ui/src/routines/row.rs` x1) — trivial closure → method-reference rewrites, no behavior change.
- `.changeset/enable-clippy-redundant-closure-ui.md` added (patch).

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo test --workspace` — 962 + 274 passing

PR #1070, opened 2026-07-07
chore(lint): enable clippy::unused_self in ui crate

tupe12334:
## What

Enables `clippy::unused_self` in `ui/Cargo.toml`'s `[lints.clippy]` table.

## Why

The `ui` (Yew/WASM) crate has its own `[lints.clippy]` table with only `all = "deny"` — it does not inherit the workspace root's extended deny-list via `[lints] workspace = true`. That means every extra lint enabled root-side (e.g. `unused_self`, denied since #1067) silently never applied to `ui/src`, even though CI's `clippy` job runs `--workspace`. This follows the same pattern as #1068 (`redundant_closure_for_method_calls` in `ui`).

The `ui` crate already has zero `unused_self` violations, so this is a lint-config-only change — no code changes, no behavior change.

## Test plan
- [x] `cargo clippy --all-targets -- -D warnings` passes for the `ui` crate

---


tupe12334:
Blocked: same as #1068 — the \`prebuilt-html-fresh\` CI check is failing, but this PR's own diff doesn't cause it. Rebuilt \`prebuilt.html\` locally from this branch and the output is byte-identical to what's already committed. Looks like toolchain/wasm-opt version drift between local and the CI runner, not a code problem this PR introduced. Needs a manual look — likely pinning the wasm toolchain versions in \`.github/workflows/prebuilt-ui.yml\`, or re-running the check.

— PR comment commander

PR #1079, opened 2026-07-07
chore(lint): enable clippy::unused_async in the ui crate

tupe12334:
## Why

`ui/Cargo.toml` has its own `[lints.clippy]` table and does not inherit the
workspace root's extended deny-list (no `[lints] workspace = true`). Root's
`Cargo.toml` denies `clippy::unused_async` — rejecting `async fn`s that never
`.await` anything, since they don't need to be async and being async
needlessly propagates async-ness to every caller — but that lint silently
never applied to `ui/src`, despite CI's `clippy` job running
`--workspace` (which lints `ui` as a member crate, just under its own,
smaller lint table).

This mirrors the same gap already fixed for `unused_self` (#1067/#1070),
`wildcard_imports` (#1073), and `unwrap_used` in `ui/Cargo.toml`, and the
in-flight #1068 for `redundant_closure_for_method_calls` — each closing the
gap for one root lint at a time. `unused_async` hadn't been picked up yet.

## What

Adds `unused_async = "deny"` to `ui/Cargo.toml`'s `[lints.clippy]` table,
with a comment mirroring the existing entries' style. The `ui` crate is
already clean under this lint (confirmed locally), so this is a pure
lock-in — no `.rs` changes needed.

Added a changeset since this touches `ui/` (`ui/Cargo.toml`).

## Checks run locally

- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean (0 new violations)
- `cargo test --workspace` — 985 + 274 passed, 0 failed
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — passes (no root `src/` changes)
- `linecheck --max-lines 500 $(find src ui/src -name '*.rs')` — clean

---


tupe12334:
Blocked: `prebuilt-html-fresh` fails non-deterministically, unrelated to this PR's diff (a zero-behavior-change lint addition). This branch already has a rebuild commit (a49b906) whose output still failed the freshness check, and re-running the same job against the same commit doesn't reliably turn it green either. Same failure reproduces on #1080. Looks like a build-environment reproducibility issue in the wasm/trunk pipeline, not something fixable by re-running or re-committing from a different machine. This check isn't in the branch ruleset's required contexts, but I won't bypass a red check to merge. Needs owner attention on the workflow itself (e.g. pin toolchain versions, build with reproducible flags, or have CI commit the regenerated file instead of verifying byte-identity across environments).

— PR comment commander

tupe12334:
Blocked: the `prebuilt-html-fresh` check fails on this PR. Same systemic toolchain/trunk-version drift seen on #1082 and #1080 — a local rebuild of `prebuilt.html` matches the committed file, so this isn't a defect in this PR's diff. Needs a maintainer with the exact CI toolchain to regenerate, or the workflow to pin trunk/rustc versions.

Failing check: `prebuilt-html-fresh` (Prebuilt UI workflow) — FAILURE.

PR #1080, opened 2026-07-07
chore(lint): enable clippy::needless_pass_by_ref_mut in the ui crate

tupe12334:
## Summary
- The `ui` crate has its own `[lints.clippy]` table that doesn't inherit the workspace root's extended deny-list, so `needless_pass_by_ref_mut` (denied root-side, see `Cargo.toml`) never applied to `ui/src` despite CI's `clippy` job running `--workspace`.
- Mirrors the lint into `ui/Cargo.toml`, following the same pattern as the crate's existing mirrored lints (`unused_self`, `wildcard_imports`, `unwrap_used`, `redundant_closure_for_method_calls`).
- The crate is already clean — this is a zero-diff lock-in, no behavior change.

## Test plan
- [x] `cargo clippy -p ui --all-targets -- -D warnings` passes
- [x] Changeset added (`patch`, no behavior change)


tupe12334:
Blocked: `prebuilt-html-fresh` fails non-deterministically, unrelated to this PR's diff (a zero-behavior-change lint addition). I regenerated `prebuilt.html` locally (`cargo check` with trunk 0.21.14, matching CI's installed version) and pushed it — CI still reports the freshly-built file differs from committed. Same failure reproduces on #1079 with an identical (already-committed) fix attempt. This looks like a build-environment reproducibility issue in the wasm/trunk pipeline (e.g. debug-info/codegen-unit ordering), not something fixable by re-running or re-committing from a different machine. Note this check isn't in the branch ruleset's required contexts, but I won't bypass a red check to merge. Needs owner attention on the workflow itself (e.g. pin toolchain versions, build with reproducible flags, or have CI commit the regenerated file instead of verifying byte-identity across environments).

— PR comment commander

tupe12334:
Blocked: the `prebuilt-html-fresh` check fails on this PR. A local rebuild of `prebuilt.html` via trunk matches the committed file byte-for-byte, so this looks like a toolchain/trunk-version drift between the CI runner and any local rebuild rather than a defect in this PR's diff (same failure seen on #1082, another ui-crate-only PR). Needs a maintainer with the exact CI toolchain to regenerate, or the workflow to pin trunk/rustc versions.

Failing check: `prebuilt-html-fresh` (Prebuilt UI workflow) — FAILURE.

Issue #241, opened 2026-06-18
Routine `repositories` entries are not validated — empty/whitespace repository URL is accepted and emits a broken clone bullet in the prompt preamble


tupe12334:
## Problem

`CreateRoutineRequest`/`UpdateRoutineRequest` accept a `repositories: Vec<Repository>` but neither `svc_create` nor `svc_update` (`src/routines/service.rs`) validates the entries. Each `Repository.repository` is a free-form `String` (`src/routines/model.rs`) and is stored verbatim, then rendered directly into the agent's `prompt.md` preamble by `compose_prompt` (`src/routines/command.rs`):

```rust
for repo in &routine.repositories {
    match &repo.branch {
        Some(branch) => body.push_str(&format!("- {} (branch {})\n", repo.repository, branch)),
        None => body.push_str(&format!("- {}\n", repo.repository)),
    }
}
```

So an empty or whitespace-only `repository` URL produces a broken bullet — `- ` (or `-  (branch x)`) — under the line *"These repositories are relevant — clone any you need:"*. The agent is handed a context list that points at nothing, and a blank/garbage URL silently survives create + update.

## Motivation

This is the same class of silent-input bug already filed for the other routine fields:

- #226 — empty/whitespace `title`
- #224 — empty/whitespace `prompt`
- #139 — unknown `agent`
- #233 — `ttl_secs` / `max_runtime_secs` of 0

`repositories` is the remaining unvalidated user-facing field. A bad entry degrades the agent run (misleading clone instruction, wasted context) instead of being rejected at the API boundary where every other field is checked.

## Acceptance criteria

- On create and update, reject a routine whose `repositories` contains any entry with an empty or whitespace-only `repository` URL, returning the existing validation error type (e.g. `AppError::BadRequest`) with a clear message — consistent with the #226/#224 validation style.
- Trim surrounding whitespace before storing, or reject leading/trailing whitespace, so `compose_prompt` never emits a malformed bullet.
- An empty `repositories` list (zero entries) remains valid — this only guards non-empty entries.
- Tests cover: empty URL rejected, whitespace-only URL rejected, valid URL accepted, empty list accepted — on both create and update paths.
- (Optional) Apply the same emptiness check to `branch` when `Some`, so `- url (branch )` cannot be produced.

---
*This issue was suggested by the "Suggest new GitHub issues (daemon + landing)" routine of moadim as a SUGGESTION. It is intentionally left unassigned — approve it by self-assigning. cc @tupe12334*

Issue #237, opened 2026-06-18
compose_prompt emits the "repositories are relevant" preamble even when a routine has no repositories


tupe12334:
## Problem

`compose_prompt` (`src/routines/command.rs`) always emits the repositories preamble, regardless of whether the routine actually has any repositories:

```rust
let mut body = String::from("# Workbench\n");
body.push_str(
    "You are working in an empty directory. These repositories are relevant — clone any you need:\n",
);
for repo in &routine.repositories {
    ...
}
```

When `routine.repositories` is empty, the generated `prompt.md` becomes:

```
# Workbench
You are working in an empty directory. These repositories are relevant — clone any you need:

---
<the prompt>
```

So a repo-less routine hands the agent a dangling header that announces "These repositories are relevant — clone any you need:" followed by **no list**. The instruction is self-contradictory (relevant repos exist, but none are named) and adds noise that can confuse the agent into looking for or cloning repositories that were never configured.

## Motivation

- Repositories are an optional field (`#[serde(default)]` on `Routine::repositories`), so repo-less routines are a fully supported, normal case — yet they get a misleading prompt.
- Prompt quality directly drives agent behavior; a contradictory preamble is a real (if low-severity) correctness wart on every repo-less run.
- Existing tests (`compose_prompt_lists_repos_and_prompt`, `compose_prompt_repo_without_branch` in `src/routines/mod_tests.rs`) only cover the non-empty case, so the empty-repos shape is untested.

## Acceptance criteria

- [ ] When `routine.repositories` is empty, `compose_prompt` omits the "These repositories are relevant — clone any you need:" line (and ideally the dangling structure), producing a clean prompt — e.g. keep `# Workbench` / "empty directory" context without the empty repo list, or drop the repo block entirely.
- [ ] When `routine.repositories` is non-empty, output is unchanged (existing tests still pass).
- [ ] Add a `compose_prompt` test covering the empty-`repositories` case asserting the misleading line is absent.

---
*This issue was opened by the "Suggest new GitHub issues (daemon + landing)" routine of moadim, as a suggestion for @tupe12334 to review. It is intentionally left unassigned — self-assign to approve.*

PR #1081, opened 2026-07-07
feat(routines): add read-only prompt-preview endpoint (#391)

### LINKED ISSUE #391: Expose a side-effect-free composed-prompt preview (GET /routines/{id}/prompt-preview + MCP tool) — today you can only see what the agent receives by triggering a real run
## Problem

The body the agent actually receives — `compose_prompt(routine)` in `src/routines/command.rs` — is only materialized when a run fires: `routine_storage.rs:150` writes it to `prompt.md` inside a fresh workbench. There is no way to **preview** that composed body without triggering a real agent run (or SSHing in to read `prompt.md` off disk). The REST API, MCP tools, and UI all expose the raw stored `routine.prompt`, never the *composed* output that prepends the `# Workbench` preamble and the per-repo clone bullets:

```
# Workbench
You are working in an empty directory. These repositories are relevant — clone any you need:
- <repo> (branch <b>)

---
<routine.prompt>
```

This is exactly the surface where a class of already-filed bugs manifests, and where the user would want to verify *before* paying for a run:
- #237 — preamble emitted even when the routine has no repositories
- #241 — blank/whitespace repository → broken `- ` clone bullet
- #262 — non-ASCII repo/title handling
- #152 / #167 — routine-origin disclosure correctness

Today the only way to confirm "what will the agent see?" is to let it fire.

## Motivation

A dry-run preview turns prompt composition into a verifiable artifact:
- Operators (and the AI agents that drive the MCP interface) can confirm the repo clone bullets, the preamble, and the final body **before** a scheduled or manual run consumes a workbench and an agent session.
- It gives every prompt-composition bug a cheap, side-effect-free reproduction surface (no tmux session, no workbench, no agent spawn).
- It mirrors the existing read-only `GET /routines/{id}` shape, just returning the *derived* body instead of the stored fields.

## Acceptance criteria

- [ ] Add a read-only `GET /routines/{id}/prompt-preview` (or `?preview=composed` on the existing route) that returns the exact `compose_prompt(routine)` body for the persisted routine, computed in-memory with **no** workbench creation, file write, or agent launch.
- [ ] Expose the same via an MCP tool (e.g. `preview_routine_prompt`) so AI callers can verify composition over MCP, consistent with the existing `get_routine` tool.
- [ ] 404 for an unknown routine id, matching `get_routine` / `svc_get` error behavior.
- [ ] Response documents whether the system-prompt / routine-origin disclosure (written separately to `CLAUDE.md`, not part of `compose_prompt`) is included; if out of scope, say so explicitly in the field/docs so callers know the preview is the `prompt.md` body only.
- [ ] OpenAPI spec + README updated for the new endpoint.
- [ ] Tests: preview matches `compose_prompt` byte-for-byte for a routine with 0 repos, 1 repo, and a branch-qualified repo; unknown id returns 404.

---
_This issue was filed as a suggestion by the "Suggest new GitHub issues (daemon + landing)" routine of moadim. It is intentionally left **unassigned** — self-assign it to approve._

tupe12334:
## What

Adds `GET /routines/{id}/prompt-preview` and the matching MCP tool `preview_routine_prompt`, both returning `compose_prompt(routine)` — the exact body an agent run would receive — computed **in-memory**, with no workbench creation, `prompt.md` write, or agent launch.

## Why

Today the composed prompt body (the `# Workbench` preamble + per-repo clone bullets, prepended to the stored `routine.prompt`) is only materialized when a run actually fires. There is no way to see what an agent will receive without triggering a real run or reading `prompt.md` off disk on the host. This is exactly the surface where a class of already-fixed prompt-composition bugs manifested (#237 preamble emitted with no repos, #241 blank-repo clone bullet, #262 non-ASCII handling) — a side-effect-free preview turns prompt composition into something operators (and MCP-driving agents) can verify *before* a scheduled or manual run consumes a workbench and an agent session.

Closes #391 (REST + MCP surface; OpenAPI/README updated). The response is the `prompt.md` body only — it does not include the routine-origin disclosure written separately to `CLAUDE.md` at trigger time, which is called out explicitly in the doc comments.

## Implementation

- `svc_get_prompt_preview` in `src/routines/service_run_files.rs` — looks up the routine and calls the existing `compose_prompt`, mirroring `svc_get`'s lookup pattern. (Placed here rather than a new module/`service.rs` itself, since `service.rs` is already at the 500-line linecheck gate.)
- `GET /routines/{id}/prompt-preview` handler + route registration, following the exact pattern of `GET /routines/{id}/logs` (plain-text response, 404 on unknown id).
- `preview_routine_prompt` MCP tool, following the `get_routine`/`routine_logs` tool pattern.
- OpenAPI spec (`apis/openapi.json`, regenerated) and README (REST + MCP tool lists) updated.
- Tests: preview matches `compose_prompt` byte-for-byte for a routine with 0 repositories and with repositories (including a branch-qualified repo), plus a 404-on-unknown-id case.

## Verification

- `cargo fmt --check` — pass
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — pass
- `cargo test --workspace` — 988 + 274 passed, 0 failed (includes the openapi spec self-heal + re-verify round trip)
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items` — pass
- `linecheck --max-lines 500` over `src/` + `ui/src/` — pass

---

### DIFF
--- a/.changeset/prompt-preview-endpoint.md
+++ b/.changeset/prompt-preview-endpoint.md
@@ -1,0 +1,5 @@
+---
+"moadim": patch
+---
+
+Add `GET /routines/{id}/prompt-preview` (and the matching `preview_routine_prompt` MCP tool) to return the exact composed prompt body a routine's run would receive, computed in-memory with no workbench, `prompt.md` write, or agent launch. Lets operators verify repo clone bullets and prompt composition before a scheduled or manual run consumes a workbench.
--- a/README.md
+++ b/README.md
@@ -206,6 +206,7 @@ PUT    /routines/{id}         # replace
 PATCH  /routines/{id}         # update fields
 DELETE /routines/{id}         # delete
 POST   /routines/{id}/trigger # run now, outside the schedule
+GET    /routines/{id}/prompt-preview # composed prompt body a run would receive, no run
 GET    /routines/{id}/logs    # run output
 POST   /routines/cleanup      # reap expired workbenches now
 GET    /agents                # list registered agents
@@ -213,8 +214,9 @@ GET    /routines.ics          # subscribe to fire times as a calendar feed
 ```
 
 **MCP** — the same operations are exposed as tools: `list_routines`,
-`get_routine`, `create_routine`, `update_routine`, `delete_routine`,
-`trigger_routine`, `routine_logs`, `list_agents`, and `cleanup_workbenches`.
+`get_routine`, `preview_routine_prompt`, `create_routine`, `update_routine`,
+`delete_routine`, `trigger_routine`, `routine_logs`, `list_agents`, and
+`cleanup_workbenches`.
 
 **Agents:** the `agent` field resolves to a config at
 `~/.config/moadim/agents/<agent>.toml`. API responses include
--- a/apis/openapi.json
+++ b/apis/openapi.json
@@ -800,6 +800,34 @@
         }
       }
     },
+    "/routines/{id}/prompt-preview": {
+      "get": {
+        "tags": [
+          "crate::routines"
+        ],
+        "summary": "`GET /routines/{id}/prompt-preview` — the exact prompt body a run would receive, computed\nin-memory with no workbench, `prompt.md` write, or agent launch (issue #391). Does not include\nthe routine-origin disclosure written separately to `CLAUDE.md` at trigger time.",
+        "operationId": "get_prompt_preview",
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "Routine UUID",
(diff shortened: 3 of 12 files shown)


tupe12334:
Blocked: `cargo llvm-cov (100% line floor)` fails — `routines/handlers.rs` is at 95.20% line coverage (11 uncovered lines), below the required 100% floor. The new prompt-preview endpoint needs additional tests to cover its branches.

Failing check: `cargo llvm-cov (100% line floor)` (Test workflow) — FAILURE.

PR #1085, opened 2026-07-07
chore(lint): enable clippy::dbg_macro/todo/unimplemented in the ui crate

tupe12334:
## Summary

The `ui` (Yew/WASM) crate has its own `[lints.clippy]` table with only `all = "deny"` — it does not inherit the workspace root's extended deny-list via `[lints] workspace = true`. Root `Cargo.toml` already denies `dbg_macro`, `todo`, and `unimplemented` to keep debugging leftovers and stub panics out of production, but because `ui` doesn't inherit that table, none of the three ever applied to `ui/src`, despite CI's `clippy` job running `--workspace`.

Concretely: a stray `dbg!()`, `todo!()`, or `unimplemented!()` left in the UI crate today would compile clean and ship straight into the release WASM bundle, panicking the running Yew app the moment that code path is hit — the exact class of failure these lints already prevent server-side.

This follows the same pattern as the last several ui-crate lint mirrors (`unused_async` in #1079, `needless_pass_by_ref_mut`, `match_same_arms`, `map_unwrap_or`, `unwrap_used`, `redundant_closure_for_method_calls`, `wildcard_imports`, `unused_self`), each closing the same root/ui inheritance gap for one lint at a time.

- Adds `dbg_macro`, `todo`, and `unimplemented` to `ui/Cargo.toml`'s `[lints.clippy]` table, with comments explaining why (mirroring root's).
- No code changes: `ui/src` already has zero violations of all three.
- Added a changeset (`patch`) per `CONTRIBUTING.md`.

## Verification

- `cargo fmt --check` — clean
- `cargo test --workspace` — 274 (ui) + 991 (root) passed, 0 failed
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — 100% lines
- `linecheck --max-lines 500` over `src/` + `ui/src/` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items` — clean

## Test plan

- [x] CI `clippy` job (runs `--workspace --all-targets`)
- [x] CI `test` job
- [x] CI `changelog`/changeset-presence check


tupe12334:
Blocked: the `prebuilt-html-fresh` CI check fails on this PR's own diff, but the diff only touches `ui/Cargo.toml`'s `[lints.clippy]` table (adding `dbg_macro`/`todo`/`unimplemented` deny attributes) — it doesn't touch `ui/src` or any code that affects the compiled UI bundle.

Locally reproducing the exact CI steps (fresh `cargo install --locked trunk` → resolves to 0.21.14, same as CI; `cargo check`/`cargo build` on this branch, on this branch merged with current `main`, and from a fully clean `target/`) always regenerates a `prebuilt.html` byte-identical to what's already committed — no diff. So the failure isn't reproducible from the PR's own content or a stale/behind-main state.

That points to the CI job itself (`.github/workflows/prebuilt-ui.yml`) producing a different artifact than a local build on the same trunk version — most likely because the committed `prebuilt.html` was last regenerated on a different OS/toolchain than the `ubuntu-latest` runner this check runs on, and `trunk build --release`'s output isn't guaranteed byte-identical across platforms (e.g. embedded paths in wasm-bindgen glue). That's an environment/reproducibility gap in the check itself, not a defect in this PR's diff — fixing it needs an owner decision (e.g. regenerate & commit `prebuilt.html` from the same Linux environment CI uses, or pin/normalize the build so it's deterministic across OSes) rather than a code change to this PR.

— PR comment commander

PR #1092, opened 2026-07-07
chore(lint): enable clippy::unreadable_literal in the ui crate

tupe12334:
## Why

`unreadable_literal` is denied workspace-root-side in `Cargo.toml`, but the `ui` (Yew/WASM) crate has its own `[lints.clippy]` table with only `all = "deny"` — it does not inherit the root's extended deny-list via `[lints] workspace = true`. As a result this lint silently never applied to `ui/src`, even though CI's `clippy` job runs `--workspace`. This is the same gap already closed for several other lints in `ui/Cargo.toml` (`unused_self`, `wildcard_imports`, `map_unwrap_or`, `match_same_arms`, `unused_async`, `uninlined_format_args`, etc).

A future 5+-digit numeric literal added to `ui/src` without `_` separators (e.g. `86400` instead of `86_400`) would currently pass CI unnoticed, hurting at-a-glance readability.

## What

- Add `unreadable_literal = "deny"` to `ui/Cargo.toml`'s `[lints.clippy]` table.
- Add a changeset (`.changeset/enable-clippy-unreadable-literal-ui.md`) — this repo requires one for any PR touching `src/` or `ui/`.

## Verification

The `ui` crate is already clean under this lint — no code changes were needed. Ran locally, matching CI/pre-push exactly:

- `cargo fmt --check` — pass
- `cargo test --workspace` — 274 passed (ui crate suite)
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — pass (root package, unaffected by this ui-only change)

No behavior change.


tupe12334:
Blocked: the `prebuilt-html-fresh` CI check fails on this PR's own diff, but the diff only touches `ui/Cargo.toml`'s `[lints.clippy]` table (adding a `clippy::unreadable_literal` deny attribute) and a changeset — it doesn't touch `ui/src` or anything that affects the compiled UI bundle.

Same root cause as sibling PR #1085: this looks like an environment/reproducibility gap in the check itself (committed `prebuilt.html` likely last regenerated on a different OS/toolchain than the `ubuntu-latest` CI runner, and `trunk build --release` output isn't guaranteed byte-identical across platforms), not a defect in this PR's diff. Fixing it needs an owner decision (e.g. regenerate & commit `prebuilt.html` from the same Linux environment CI uses, or make the build deterministic across OSes) rather than a code change to this PR.

— PR comment commander

PR #1082, opened 2026-07-07
chore(lint): enable clippy::map_unwrap_or in the ui crate

tupe12334:
## Summary

The root crate already denies `clippy::map_unwrap_or` (`Cargo.toml`), rejecting the
`.map(f).unwrap_or(_)` / `.map(f).unwrap_or_else(_)` pattern in favour of the single-combinator
`map_or`/`map_or_else`/`is_some_and` forms — clearer intent, no throwaway closure allocation for
the "already produced a value" path.

That deny-list doesn't reach the `ui` crate: `ui/Cargo.toml` has its own `[lints.clippy]` table
with no `workspace = true` inheritance, so `ui/src` silently escaped this lint despite CI's
`clippy` job running with `--workspace`. This is the same gap PRs #1079 and #1080 are closing for
`unused_async` and `needless_pass_by_ref_mut` — bringing `ui` up to parity with the root crate's
extended lint set one lint at a time.

- Adds `map_unwrap_or = "deny"` to `ui/Cargo.toml`'s `[lints.clippy]` table.
- Fixes the 7 violations this surfaced:
  - `log_viewer.rs` — `.map(String::len).unwrap_or(0)` -> `.map_or(0, String::len)`
  - `overview_recent_runs.rs` (2x) — `.map(f).unwrap_or_else(|| "—".to_string())` -> `.map_or_else(|| "—".to_string(), f)`
  - `routines/form.rs` — same `map_or_else` rewrite for `agent`, plus `.map(|r| r.enabled).unwrap_or(true)` -> `.is_none_or(|r| r.enabled)` (clippy's `unnecessary_map_or`, part of `clippy::all`, prefers this over `map_or(true, ...)`)
  - `routines/history.rs` — same `map_or_else` rewrite for `exit_code`
  - `routines/hooks.rs` — `.map(|el| ...).unwrap_or(false)` -> `.is_some_and(|el| ...)`
- Regenerates `prebuilt.html` per the `ui/`-change convention in CONTRIBUTING.md.

All rewrites are behavior-preserving.

## Verification

- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test --workspace` — 985 + 274 passed, 0 failed
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — 100% lines (exit 0)
- `linecheck --max-lines 500 $(find src ui/src -name '*.rs')` — clean

PR #1110, opened 2026-07-10
chore(lint): enable clippy::needless_collect in the ui crate

tupe12334:
## Summary

- Adds `needless_collect = "deny"` to `ui/Cargo.toml`'s `[lints.clippy]` table, mirroring the root crate's `needless_collect = "deny"` (`Cargo.toml`).
- The `ui` crate has its own `[lints.clippy]` table with no `workspace = true` inheritance, so this lint (like several others before it — #1109, #1108, #1106, #1102, #1085, #1091, #1083, #1082, #1080, #1079, #1078, #1076, #1073, #1070, #1068) silently never applied to `ui/src` despite CI's `clippy` job running `--workspace`.
- `needless_collect` catches an iterator collected into a `Vec`/collection only to be immediately iterated again (or have its length/emptiness checked) — the intermediate allocation is pure overhead the chain never needed. It's a real perf-adjacent gap, not just a style one.
- The `ui` crate is already clean under this lint, so no code changes are needed — `deny` just locks that state in.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --workspace` (999 root + 280 ui tests, all passing)
- [x] Added `.changeset/enable-clippy-needless-collect-ui.md` (patch, package `moadim`)


tupe12334:
Blocked: `prebuilt-html-fresh` fails because the wasm bundle isn't byte-reproducible across build environments — I regenerated `prebuilt.html` locally (matching CI's rustc 1.97.0) and it picks up the real content drift (e.g. the sparkline CSS/JS already in `ui/src` but missing from the committed bundle), but the wasm binary itself still differs byte-for-byte from a fresh CI build, unrelated to this PR's diff (only `ui/Cargo.toml` lint config changed). Reproducing CI's exact container here isn't available (no docker network egress in this environment). This needs either a deterministic build fix (pinned trunk/wasm-bindgen version + `--remap-path-prefix`) or a maintainer rebuilding `prebuilt.html` from the exact CI image.

PR #1112, opened 2026-07-10
chore(lint): enable clippy::format_push_string in the ui crate

tupe12334:
## Why

`format_push_string` is denied workspace-root-side in `Cargo.toml`, but the `ui` (Yew/WASM) crate has its own `[lints.clippy]` table with only a handful of lints mirrored over — it does not inherit the root's extended deny-list via `[lints] workspace = true`. As a result this lint silently never applied to `ui/src`, even though CI's `clippy` job runs `--workspace`. This is the same gap already closed for several other lints in `ui/Cargo.toml` (`unused_self`, `wildcard_imports`, `map_unwrap_or`, `match_same_arms`, `unused_async`, `dbg_macro`, `todo`/`unimplemented`, `uninlined_format_args`, `unnecessary_debug_formatting`, `case_sensitive_file_extension_comparisons`, `trivially_copy_pass_by_ref`, `redundant_else`, and in-flight `needless_collect` (#1110) / `unreadable_literal` (#1092)).

`format_push_string` catches `.push_str(&format!(...))` — the `format!` call allocates a throwaway `String` only to immediately copy its contents into the target and drop it, so it's pure overhead that `write!`/`writeln!` avoids by writing straight into the existing buffer. A real perf-adjacent gap, not just a style one.

## What

- Add `format_push_string = "deny"` to `ui/Cargo.toml`'s `[lints.clippy]` table.
- Add a changeset (`.changeset/enable-clippy-format-push-string-ui.md`) — this repo requires one for any PR touching `src/` or `ui/`.

## Verification

The `ui` crate is already clean under this lint — no code changes were needed. Ran locally, matching CI/pre-push exactly:

- `cargo fmt --check` — pass
- `cargo test --workspace` — 999 root + 280 ui tests, all passing
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — pass (root package, unaffected by this ui-only change)
- `linecheck --max-lines 500` — pass
- `./scripts/check-readme-blocks.sh` — pass

No behavior change.

### DIFF
--- a/.changeset/enable-clippy-format-push-string-ui.md
+++ b/.changeset/enable-clippy-format-push-string-ui.md
@@ -1,0 +1,17 @@
+---
+"moadim": patch
+---
+
+chore(lint): enable `clippy::format_push_string` in the `ui` crate
+
+Adds `format_push_string = "deny"` to `ui/Cargo.toml`'s `[lints.clippy]` table, mirroring the
+root crate's `format_push_string = "deny"` (`Cargo.toml`). The `ui` crate has its own
+`[lints.clippy]` table with no `workspace = true` inheritance, so this lint (like several others
+before it) silently never applied to `ui/src` despite CI's `clippy` job running `--workspace`.
+
+`format_push_string` catches `.push_str(&format!(...))`, which allocates a throwaway `String`
+only to immediately copy its contents into the target and drop it — a real perf-adjacent gap, not
+just a style one. `write!`/`writeln!` write straight into the existing buffer instead.
+
+The `ui` crate is already clean under this lint, so no code changes are needed — `deny` just locks
+that state in. No behavior change.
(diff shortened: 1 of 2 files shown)


tupe12334:
Blocked: the `prebuilt-html-fresh` CI check fails on this PR's own diff, but the diff only touches `ui/Cargo.toml`'s `[lints.clippy]` table (adding a `clippy::format_push_string` deny attribute) and a changeset — it doesn't touch `ui/src` or anything that affects the compiled UI bundle.

Same root cause as sibling PRs #1092 and #1110: this is an environment/reproducibility gap in the check itself (committed `prebuilt.html` isn't byte-reproducible across build environments), not a defect in this PR's diff. Fixing it needs an owner decision (e.g. regenerate & commit `prebuilt.html` from the exact CI image, or make the build deterministic) rather than a code change to this PR.

— PR comment commander

PR #1113, opened 2026-07-10
chore(lint): enable clippy::single_match_else in the ui crate

tupe12334:
## Why

The root crate denies `clippy::single_match_else` (see `Cargo.toml`), but the `ui` crate has its
own `[lints.clippy]` table with no `workspace = true` inheritance, so this lint has silently
never applied to `ui/src` despite CI's `clippy` job running `--workspace`. This continues the
existing pattern of mirroring root-side lints into `ui/Cargo.toml` one at a time (see #1112,
#1109, #1108, #1106, #1085, #1092), picking a lint not already covered by another open PR
(`needless_collect` is proposed separately in #1110).

`single_match_else` catches a `match` whose only non-wildcard arm destructures a single pattern,
with the rest falling to a catch-all `else`-like arm — `if let ... else` says the same thing
without the unused generality of `match`.

## What

Adds `single_match_else = "deny"` to `ui/Cargo.toml`. Verified the `ui` crate is already clean
under this lint, so no source changes are needed — this only locks in the current clean state.
No behavior change.

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [x] `cargo test --workspace` passes (999 + 280 tests)
- [x] Added `.changeset/enable-clippy-single-match-else-ui.md`


tupe12334:
Blocked: resolved the ui/Cargo.toml merge conflict with main (additive lint-table entries, kept both), fmt/clippy/build/test all pass, branch pushed. prebuilt-html-fresh still fails though - same build-environment reproducibility issue as PR #1122/#1120/#1119 (non-deterministic wasm codegen/asset hashing), not a real content problem.

— PR comment commander

PR #1117, opened 2026-07-10
chore(lint): enable clippy::manual_string_new in the ui crate

tupe12334:
## Why

Following the established pattern of locking in one clippy lint at a time for the `ui` crate (see #1116, #1115, #1113, #1110, etc.): `manual_string_new` is already `deny`d workspace-root-side in `Cargo.toml`, but the `ui` crate has its own `[lints.clippy]` table with no `workspace = true` inheritance, so it silently escaped this lint despite CI's `clippy` job running `--workspace`.

## What

Adds `manual_string_new = "deny"` to `ui/Cargo.toml`. The crate is already clean under this lint — no code changes needed, this just locks the state in so a future violation fails CI.

## Verification

- `cargo fmt --check` — clean
- `cargo test -p ui` — 280 passed

### DIFF
--- a/.changeset/enable-clippy-manual-string-new-ui.md
+++ b/.changeset/enable-clippy-manual-string-new-ui.md
@@ -1,0 +1,11 @@
+---
+"moadim": patch
+---
+
+chore(lint): enable `clippy::manual_string_new` in the `ui` crate
+
+Adds `manual_string_new = "deny"` to `ui/Cargo.toml`'s `[lints.clippy]` table, matching the
+lint already denied workspace-root-side in `Cargo.toml` — the `ui` crate has its own `[lints]`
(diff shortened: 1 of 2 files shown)


tupe12334:
Blocked: `prebuilt-html-fresh` fails — same known issue as #1110. The committed `prebuilt.html` doesn't match a fresh `cargo build` output on CI's toolchain (rustc 1.97.0); this environment's rustc (1.96.1, macOS) can't reproduce CI's Linux/wasm32 build byte-for-byte to confirm whether the diff is content drift or pure non-determinism. This PR's diff is lint-config only (`ui/Cargo.toml`), so the mismatch is pre-existing drift, not something introduced here. Needs either a deterministic build fix (pinned trunk/wasm-bindgen + `--remap-path-prefix`) or a maintainer rebuilding `prebuilt.html` from the exact CI image.

— PR comment commander

PR #1119, opened 2026-07-11
chore(lint): enable clippy::semicolon_if_nothing_returned in the ui crate

tupe12334:
## Why

The `ui` crate has its own `[lints.clippy]` table (no `workspace = true` inheritance), so several lints already denied workspace-root-side in `Cargo.toml` never actually applied to `ui/src`, despite CI's `clippy` job running `--workspace` — the lint config itself was the gap, not the code. This repo has been closing that gap one lint at a time (see the many merged `chore(lint): enable clippy::X in the ui crate` PRs); this PR does the same for `clippy::semicolon_if_nothing_returned`.

`semicolon_if_nothing_returned` flags a block whose final expression evaluates to `()` but omits the trailing `;` — the block reads as if it's returning that expression's value in tail position, when it's actually just a statement whose result is discarded. That's a real (if minor) readability trap in exactly the kind of event-handler code this UI is full of.

## What

- Add `semicolon_if_nothing_returned = "deny"` to `ui/Cargo.toml`.
- Fix the 10 violations this surfaced: `Callback::from` closures in `ui/src/main.rs`, `ui/src/routines/actions.rs`, and `ui/src/routines/page.rs` whose body was a bare `spawn_local(...)` or `toast.emit(...)` call with no trailing semicolon. Applied via `cargo clippy --fix`, so each is a mechanical added `;` — no behavior change.
- Regenerate `prebuilt.html` so the committed fallback bundle stays in lockstep with `ui/src` (enforced by the `prebuilt-html-fresh` CI job).
- Add a changeset (`.changeset/enable-clippy-semicolon-if-nothing-returned-ui.md`) per the repo's changesets migration.

## Test plan

- [x] `cargo fmt --check` — passes
- [x] `cargo clippy -p ui --all-targets -- -D warnings` — passes (was 10 errors before the fix)
- [ ] CI: `cargo clippy --workspace --all-targets -- -D warnings` (this repo's sandboxed dev environment hit disk-space limits on the full workspace build during this session, so the workspace-wide run relies on CI; the `-p ui` run above covers every file this PR touches)


tupe12334:
Blocked: prebuilt-html-fresh fails, same root cause as PR #1122/#1120 - build-environment reproducibility issue in the freshness check (non-deterministic wasm codegen/asset hashing), not a real content problem. Branch synced to main; all other checks pass.

— PR comment commander

PR #1120, opened 2026-07-11
chore(lint): enable clippy::explicit_iter_loop in the ui crate

tupe12334:
## Why

The root crate denies `clippy::explicit_iter_loop` (`Cargo.toml`), but the `ui` crate has its own `[lints.clippy]` table with no `workspace = true` inheritance — so this lint silently never applied to `ui/src` despite CI's `clippy` job running `--workspace`. This is the same gap already fixed piecemeal by the sibling `chore(lint): enable clippy::X in the ui crate` PRs.

Unlike most of those, this one isn't a no-op switch-flip: it surfaced two real violations in `day_timeline.rs`.

## What

- Add `explicit_iter_loop = "deny"` to `ui/Cargo.toml`'s `[lints.clippy]` table.
- Fix the two violations it surfaced in `ui/src/day_timeline.rs`:
  - `for it in props.items.iter()` → `for it in &props.items`
  - `for b in buckets.iter_mut()` → `for b in &mut buckets`
- Add a changeset (patch bump, no behavior change).

Verified locally with `cargo clippy -p ui --all-targets -- -D warnings` (clean).

No behavior change.

### DIFF
--- a/.changeset/enable-clippy-explicit-iter-loop-ui.md
+++ b/.changeset/enable-clippy-explicit-iter-loop-ui.md
@@ -1,0 +1,13 @@
+---
+"moadim": patch
+---
+
+chore(lint): enable `clippy::explicit_iter_loop` in the `ui` crate
+
+Adds `explicit_iter_loop = "deny"` to `ui/Cargo.toml`'s `[lints.clippy]` table, matching the lint
+already denied workspace-root-side in `Cargo.toml` — the `ui` crate has its own `[lints]` table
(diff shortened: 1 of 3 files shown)


tupe12334:
Blocked: prebuilt-html-fresh fails, same as PR #1122 - this appears to be a build-environment reproducibility issue in the freshness check itself (non-deterministic wasm codegen/asset hashing), not a real content problem in prebuilt.html vs ui/ source. Not something auto-fixable via changeset/sync/conflict-resolution. Branch was synced to main; all other checks pass.

— PR comment commander

PR #1122, opened 2026-07-11
fix(build): regenerate stale prebuilt.html

tupe12334:
## Why

`prebuilt.html` — the self-contained UI bundle `build.rs` embeds at compile
time as a fallback for builds without `trunk` (e.g. `cargo install moadim`
from git, or a Docker build that skips trunk) — hasn't been regenerated
since #1092. Every merge to `ui/src` since then recompiles the embedded
JS/WASM bytes, so the `prebuilt-html-fresh` CI job now fails on any PR that
touches `ui/`, even ones with no visual/behavioral change.

This is currently blocking several open lint-enablement PRs (#1119, #1120,
#1113) purely on this drift, not on their own content.

## What

Rebuilt `prebuilt.html` locally via `cargo check` (which runs `build.rs` /
`trunk`, same as CI's `prebuilt-html-fresh` job) and committed the result.
No source changes — binary bundle only.

## Test plan
- [x] `cargo check` regenerates `prebuilt.html` with no further diff (fixed point)
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] Added `.changeset/regen-stale-prebuilt-html.md`

---


tupe12334:
Blocked: prebuilt-html-fresh still fails after merging main and regenerating prebuilt.html via trunk build --release (the documented fix). The rebuilt bytes differ from what CI own rebuild expects, most likely non-deterministic wasm codegen/asset hashing between build environments rather than a content issue - ui/ source and prebuilt.html are otherwise in sync. Not a normal conflict or changeset gap; needs a look at whether the freshness check can be made environment-stable, or a manual regen from a matching CI runner.

— PR comment commander

PR #1125, opened 2026-07-12
chore(lint): enable clippy::mem_forget in the ui crate

tupe12334:
## Why

The `ui` crate has its own `[lints.clippy]` table (no `workspace = true` inheritance), so `clippy::mem_forget` — denied workspace-root-side since #1121 — silently never applied to `ui/src`, despite CI's `clippy` job running `--workspace`. Same class of gap as #1113/#1119/#1120.

## What

- `ui/Cargo.toml`: add `mem_forget = "deny"` to `[lints.clippy]`.
- `prebuilt.html`: regenerated via `trunk` to stay in lockstep with the `ui/Cargo.toml` change (`prebuilt-html-fresh` CI gate).
- `.changeset/enable-clippy-mem-forget-ui.md`: patch changeset.

No behavior change — the `ui` crate is already clean under this lint; `deny` just locks that state in.

## Test plan
- [x] `cargo fmt --all -- --check`
- [x] `cargo test --workspace`
- [x] `linecheck --max-lines 500`
- [x] `cargo check` twice in a row confirms `prebuilt.html` is stable/fresh


tupe12334:
Blocked: resolved the merge conflict with main (additive `ui/Cargo.toml` lint-table entries, kept both; `prebuilt.html` regenerated), fmt/clippy/build/test all pass, branch pushed. `prebuilt-html-fresh` still fails — same build-environment reproducibility issue as PR #1129 (confirmed root cause: the wasm binary embeds this-machine's absolute toolchain/registry paths, e.g. `/Users/...` vs CI's `/home/runner/...`, so no local rebuild — even with the pinned trunk 0.21.14 — can byte-match CI's own rebuild). Not a real content problem; needs either a CI-runner-matching rebuild or `--remap-path-prefix` in the wasm build to make the freshness check environment-stable.

— PR comment commander

PR #1127, opened 2026-07-12
fix(ci): pin trunk version so prebuilt-html-fresh stops flaking

tupe12334:
## Why

`prebuilt-html-fresh` (added in #1096-era `prebuilt-ui.yml`) rebuilds the UI with `trunk` and fails if the result doesn't byte-match the committed `prebuilt.html`. Both `prebuilt-ui.yml` and `publish.yml` install trunk with `cargo install --locked trunk` — **no version pin** — so every CI run installs whatever is *currently latest* on crates.io.

Trunk's own output (minification, asset content-hashing) isn't guaranteed byte-stable release to release. The result: the moment upstream trunk ships a new version, `prebuilt-html-fresh` starts failing on **every** open PR that touches `ui/**`, even ones that make no visible change — e.g. #1113, #1119, #1120, #1122, #1125 are all currently red on this check despite each being a scoped, source-verified change (clippy lint enablement, or a `prebuilt.html` regen). #1122 in particular regenerates `prebuilt.html` but doesn't pin the tool version, so it will go stale again the next time trunk publishes.

Confirmed locally: a clean `cargo check` from `main` (which drives `build.rs` → `trunk build --release`) produces a `prebuilt.html` that differs from what's committed, using whatever trunk is installed locally (0.21.14) — even though `ui/src` hasn't changed since the last regen. Two independent clean rebuilds against the *same* pinned trunk version, however, produce byte-identical output.

## What

- Pin `trunk` to `0.21.14` in `prebuilt-ui.yml` and `publish.yml` (`cargo install --locked --version 0.21.14 trunk`), with a comment explaining why an unpinned install is unsafe here.
- Regenerate `prebuilt.html` against that exact pinned version.
- Update `CONTRIBUTING.md`'s prerequisites table and the `prebuilt.html` regen instructions to reference the pinned version, so a contributor's local regen matches CI instead of drifting ahead of it.

No `src/` or `ui/` files changed, so the Changelog gate's `src/**`/`ui/**` filter doesn't require a changeset here (verified locally: `pnpm exec changeset status` path-filter logic and the local pre-push hook's equivalent check both pass without one).

## Verification

- `cargo fmt --all -- --check`
- `actionlint .github/workflows/prebuilt-ui.yml .github/workflows/publish.yml`
- Full `.githooks/pre-push` (test-file convention, fmt, clippy, `cargo test --workspace`, 100% line coverage, `linecheck`, changelog gate) — all green
- Two independent clean (`rm -rf target/wasm32-unknown-unknown ui/dist && cargo check`) rebuilds against the pinned trunk version produce byte-identical `prebuilt.html`

## Note for reviewers

This is a root-cause fix for the flakiness #1122 is separately working around with a one-off regen — the two will conflict on `prebuilt.html`; whichever lands first, the other should rebase onto it. Once this merges, #1113/#1119/#1120/#1122/#1125 should go green on `prebuilt-html-fresh` after a rebase/retrigger.


tupe12334:
Update: the first commit's regenerated \`prebuilt.html\` was built with my local rustc 1.96.1, but this repo's CI installs \`rustc: stable\` unpinned — and by the time this PR's checks ran, that resolved to 1.97.0 (shipped 2026-07-07, after my local toolchain). \`prebuilt-html-fresh\` failed here too, for the same class of reason this PR exists: build output isn't guaranteed byte-stable across compiler versions any more than across trunk versions.

Pushed a follow-up commit regenerating against rustc 1.97.0 (matching CI's exact commit hash `2d8144b78`), reconfirmed deterministic across two clean rebuilds under that toolchain.

Note this doesn't fully close the door on future staleness: `stable` is intentionally unpinned across this repo's CI (same as `lint.yml`/`test.yml`), so a future rustc point release can still flip this check red again, independent of trunk. Pinning rustc for just this one job to avoid that would be inconsistent with the rest of the project's CI strategy, so I left it alone — flagging it here in case a maintainer wants to make that call explicitly (e.g. a scheduled job that rebuilds prebuilt.html and opens a PR whenever it drifts, instead of a hard gate that blocks unrelated PRs).

PR #1129, opened 2026-07-12
chore(lint): enable clippy::explicit_into_iter_loop in the ui crate

tupe12334:
## Summary

- Adds `explicit_into_iter_loop = "deny"` to `ui/Cargo.toml`'s `[lints.clippy]` table, matching the lint already denied workspace-root-side in the root `Cargo.toml`.
- The `ui` crate has its own `[lints.clippy]` table (no `workspace = true` inheritance), so this lint silently never applied to `ui/src` despite CI's `clippy` job running `--workspace`.
- Direct companion to the just-merged `explicit_iter_loop` (#1120): that one catches `.iter()`/`.iter_mut()` in a `for` loop, this one catches the redundant `.into_iter()` form.
- `ui/src` has no violations today, so this is a no-op lock-in — no behavior change.

## Why

Continues the established series of PRs (#1120, #1119, #1113, #1125, etc.) that bring the `ui` crate's `[lints.clippy]` table up to parity with the root crate's, closing gaps that `cargo clippy --workspace` silently leaves open due to `ui` not inheriting the root's extended deny-list.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (passes clean with the new lint active)
- [x] `cargo test --workspace` (1000 + 280 tests pass)
- [x] Added `.changeset/enable-clippy-explicit-into-iter-loop-ui.md`


tupe12334:
Blocked: attempted the documented fix — regenerated `prebuilt.html` locally with the CI-pinned `trunk 0.21.14` (matching `.github/workflows/prebuilt-ui.yml`) and pushed it, but `prebuilt-html-fresh` still fails with a fresh byte diff on CI's own rebuild. This confirms it's a build-environment reproducibility issue (wasm codegen/asset hashing differs between this local machine and the `ubuntu-latest` runner, even with rustc/trunk pinned), not a content problem — `ui/` source and the committed `prebuilt.html` are otherwise in sync. Needs a rebuild from a matching Linux CI-like environment (no Docker daemon available here to reproduce it) or a look at making the freshness check environment-stable.

— PR comment commander

PR #1137, opened 2026-07-13
chore(lint): enable clippy::if_not_else in the ui crate

tupe12334:
## Why

The root crate denies `clippy::if_not_else` workspace-wide (see `Cargo.toml`), but the `ui` crate has its own `[lints.clippy]` table with no `workspace = true` inheritance — so this lint (like several others enabled in prior PRs) never actually applied to `ui/src` despite CI's `clippy` job running `--workspace`.

## What

- Adds `if_not_else = "deny"` to `ui/Cargo.toml`, mirroring the root crate.
- Fixes the 2 violations this surfaced:
  - `ui/src/header.rs`: the version-title span's `if !version_title.is_empty() { A } else { B }` → `if version_title.is_empty() { B } else { A }`.
  - `ui/src/overview.rs`: the attention panel's same pattern, inverted the same way.
- No behavior change — purely eliminates double-negation for readability.
- Adds a changeset (`patch`).

## Verified locally
- `cargo fmt --all -- --check`
- `cargo test --workspace` (284 passed)


tupe12334:
Merge conflict resolved (normal merge commit from origin/main — trivial additive conflict in ui/Cargo.toml's `[lints.clippy]` table). `cargo fmt`/`clippy --workspace`/`test --workspace` all pass on the merged tree.

Blocked on merging though: CI check `prebuilt-html-fresh` fails. Same systemic issue as #1143/#1142/#1140/#1138 — the trunk/wasm-bindgen build doesn't appear to be byte-reproducible across runs/environments, so this can't be fixed by mechanically re-regenerating `prebuilt.html`. Needs a maintainer call on the build/check itself.

— PR comment commander

PR #1138, opened 2026-07-13
chore(lint): enable clippy::needless_pass_by_value in the ui crate

tupe12334:
## Why

The `ui` crate has its own `[lints.clippy]` table and doesn't inherit the root crate's extended deny-list, so `clippy::needless_pass_by_value` (already `deny`d root-side, see `Cargo.toml`) never applied to `ui/src` despite CI's `clippy` job running `--workspace`. This is the same lint-parity gap already closed for `unused_self`, `map_unwrap_or`, `match_same_arms`, `semicolon_if_nothing_returned`, `manual_let_else`, and several others in past PRs.

## What

Enabling the lint surfaced 5 violations, all in the routines list page's action-installer functions:

- `install_crud_handlers` (`ui/src/routines/actions.rs`) took `state: UseReducerHandle<RState>` and `toast: Callback<(String, ToastKind)>` by value, but the body only ever `.clone()`s them into closures — the outer parameters themselves are never consumed.
- `install_bulk_handlers` (`ui/src/routines/bulk_actions.rs`) had the same shape for `state`, `toast`, and `now: UseStateHandle<DateTime<Local>>`.

Changed all five parameters to references and updated the single call site in `ui/src/routines/page.rs` to pass borrows instead of pre-cloning before the call — removing the needless ownership transfer (and the now-redundant clone at the call site) with no behavior change.

`prebuilt.html` is regenerated via `trunk` (pinned 0.21.14, matching `prebuilt-ui.yml`) to stay in lockstep with the `ui/src` change.

## Checks run locally

- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test --workspace` — 1010 (root) + 284 (ui) passed
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — 100% lines (unchanged, no `src/` touched)
- `linecheck --max-lines 500` — passed
- `.githooks/pre-push` (full gate, including the changeset check) — passed on push

A changeset is included (`.changeset/enable-clippy-needless-pass-by-value-ui.md`), `patch` bump.

This pull request was opened by the "Daemon + Landing auto-improve & auto-merge lint/docs PRs" routine of moadim. @tupe12334 feel free to tag in if you'd like a closer look before merging.

PR #1140, opened 2026-07-13
chore(lint): enable clippy::derive_partial_eq_without_eq in the ui crate

tupe12334:
## Why

The `ui` crate has its own `[lints.clippy]` table and doesn't inherit the root crate's extended deny-list (see the many prior `chore(lint): enable clippy::* in the ui crate` PRs). `derive_partial_eq_without_eq` is denied root-side but was silently exempt for `ui/src` despite CI's `clippy` job running `--workspace`.

## What

Enables `derive_partial_eq_without_eq = "deny"` in `ui/Cargo.toml` and fixes the 15 violations it surfaced by adding `Eq` alongside `PartialEq` on the affected structs/enums. All are field-for-field `Eq`-safe (no float fields) — verified by `cargo clippy --workspace --all-targets -- -D warnings` passing clean. No behavior change.

Also regenerated `prebuilt.html` locally via the pinned `trunk` 0.21.14 build (`cargo check`), since it embeds the `ui` crate's compiled output.

**Known caveat:** despite that regeneration, `prebuilt-html-fresh` still fails in CI — my local rebuild is on macOS/aarch64, and it isn't byte-identical to what `ubuntu-latest`'s `trunk`/`wasm-bindgen` toolchain produces from the same source (likely a toolchain-version or platform-dependent codegen difference upstream of this PR, not something this diff introduces). This is the same failure mode already stuck on #1137, #1138, and #1139 — a genuine rebuild from a Linux CI runner (or a maintainer's `cargo build` there) should resolve it; I don't have a Linux/Docker environment available to reproduce it byte-for-byte here.

Local `.githooks/pre-push` (fmt, clippy, workspace tests, 100% line coverage, linecheck, changeset check) all pass.

---


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own regenerated `prebuilt.html`. Same systemic issue tracked on #1143/#1142 — the trunk/wasm-bindgen build doesn't appear to be byte-reproducible across runs/environments, so a fresh CI rebuild diverges even from a correctly-regenerated commit. Not something a mechanical re-regenerate-and-commit can fix; needs a maintainer call on the build/check itself.

— PR comment commander

tupe12334:

— PR comment commander

PR #1124, opened 2026-07-11
feat(ui): saved views for the Routines page

tupe12334:
Closes #1123

## What

The Routines page's filter/sort/group-by state now persists:

- **Auto-restore**: the current filter, sort column/direction, and group-by dimension are saved to `localStorage` on every change and silently restored on the next load — a reload no longer drops an in-progress triage view.
- **Named saved views**: a new VIEWS bar (below the filter bar) lets an operator save the current filter/sort/group-by combo under a name, pick it from a dropdown to re-apply instantly, and delete views no longer needed. Also persisted to `localStorage`.

Matches the saved/pinned-view pattern used by dashboards an operator revisits repeatedly (Linear, GitHub Issues, Grafana) — see the rationale in #1123.

## Scope

Frontend-only, confined to `ui/`:
- `ui/src/routines/saved_views.rs` (new) — pure snapshot codec (`ViewSnapshot::capture`/`decode`, host-tested) + the `SavedViewsBar` component + the `localStorage` round-trip.
- `ui/src/routines/state.rs` — `RCol`/`RDir` string codecs, an `ApplySnapshot` action, `RState::init()` restoring the last view on mount. Also stops `CurrentMachineLoaded` from clobbering an already-set machine filter (e.g. one just restored from a saved view) — it now only defaults the facet when nothing has picked one yet.
- `ui/src/routines/page.rs` — wiring: loads/saves the named-view list, persists the last-used snapshot on change, renders `<SavedViewsBar>`.

No backend, API, or data-model changes.

## Test plan

- [x] `cargo build` succeeds
- [x] `cargo clippy --workspace --all-targets` clean (ui crate's `all = "deny"`)
- [x] `cargo test --workspace` — 999 (root) + 284 (ui, incl. new `saved_views_tests.rs`) passing
- [x] `cargo fmt --check` clean
- [x] `linecheck --max-lines 500` clean
- [x] `git diff --name-only origin/main...HEAD` touches only `ui/` files

PR #1142, opened 2026-07-13
fix(build): regenerate stale prebuilt.html

tupe12334:
## Summary
- \`prebuilt.html\` last regenerated at #1122 has drifted from \`ui/\` sources again through subsequent merges (#1129, #1136, #1124 landed \`ui/src\` changes after it). \`main\` currently fails its own \`prebuilt-html-fresh\` CI check as a result, and every open PR touching \`ui/\` (e.g. #1140, #1138, #1137) inherits that failure regardless of its own diff.
- Rebuilt via \`cargo check\` with trunk 0.21.14 (matching the workflow's pin) and committed the fresh bundle. No source change.

## Test plan
- [x] \`cargo check\` locally regenerates \`prebuilt.html\` with no further diff (freshness confirmed)
- [ ] CI \`prebuilt-html-fresh\` job passes


tupe12334:
Blocked: CI check `prebuilt-html-fresh` keeps failing here even though this PR's whole purpose is fixing that staleness — 3 failed runs in a row, each against a fresh regeneration, each producing a different diff. That's consistent with the trunk/wasm-bindgen build not being byte-reproducible (across runs/environments), not an actual staleness bug in this PR's diff. See #1143 for a same-commit CI rerun that reproduced identically on Linux but didn't match a local macOS rebuild of the same source — same underlying issue.

Re-regenerating again won't converge on its own; needs a maintainer call (build in the exact CI container, or relax the check to tolerate non-deterministic bytes).

— PR comment commander

tupe12334:

— PR comment commander

PR #1143, opened 2026-07-13
feat(ui): day-detail popover on the routines calendar

tupe12334:
## Why

`TODO.md` asked for exactly this: *"Add a day-detail popover to the routines calendar: clicking a day lists each fire time (HH:MM) with its routine, and a 'run now' shortcut per routine."* Today the month-view calendar only shows a handful of truncated title chips per day ("+N more" once a cell gets busy) with no way to see the full list or act on it without switching to the Routines table.

## What

- Clicking a day's date number in the routines calendar now opens a popover listing every enabled routine that fires that day, as `HH:MM — title`, sorted chronologically.
- Each row has a **▶ RUN** button that calls the existing `POST /api/v1/routines/{id}/trigger` endpoint (reuses the same `on_trigger` handler already wired to the Routines table — no new API).
- New pure helpers, both host-tested (no wasm dependency):
  - `fires_on_day(schedule, date) -> Vec<String>` in `ui/src/schedule.rs` — every fire time on a given calendar day.
  - `day_fire_rows(routines, date) -> Vec<(id, title, "HH:MM")>` in `ui/src/routines/calendar.rs` — merges and sorts across all routines for the popover.
- Removed the completed item from `TODO.md` and added the natural follow-up (extending the same popover to the Cron Jobs calendar), per the file's own convention.
- `prebuilt.html` regenerated via the pinned `trunk` 0.21.14 (matching `prebuilt-ui.yml`) to stay in lockstep with the `ui/src` change.

Frontend-only — no backend, API, or data-model changes.

## Checks run locally (`.githooks/pre-push`, full gate)

- Test-file convention (no inline `#[cfg(test)]` blocks) — passed
- `cargo fmt --all -- --check` — clean
- `cargo test --workspace` — 1010 (root) + 292 (ui, +8 new) passed
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — 100% (root package scope, unchanged; `ui/` isn't held to this floor per the hook's own comment)
- `linecheck --max-lines 500` — passed
- Changeset check — passed (`.changeset/routine-calendar-day-detail-popover.md`, `patch`)

**Known caveat:** per the pattern already seen on #1140/#1142, a locally-rebuilt `prebuilt.html` (macOS/aarch64 `trunk`) isn't guaranteed byte-identical to what the `ubuntu-latest` CI runner's `trunk`/`wasm-bindgen` toolchain produces from the same source, so `prebuilt-html-fresh` may still flag a diff in CI even though the source-level change is complete and correct. This is a pre-existing, platform-dependent CI quirk unrelated to this diff (see #1142's PR body and issue-shaped writeups on the same job) — a rebuild from a Linux runner would resolve it.

---


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails and can't be resolved by a mechanical regenerate-and-commit.

Reran the same commit's failed job with no source change — it failed again with the exact same diff hash (`5028a5f..f48dca5`), so CI itself is deterministic run-to-run. But rebuilding locally (macOS, same pinned trunk 0.21.14 + stable rustc) reproduces the *currently committed* `prebuilt.html` byte-for-byte, not the CI-built one. And PR #1142 — whose only purpose is regenerating this file — has now failed this same check 3 times in a row, each attempt producing a different diff against the last. That points to the trunk/wasm-bindgen build not being reproducible across environments/runs in general, not a staleness problem specific to this PR's diff.

Regenerating again from a non-Linux-CI environment won't converge — needs a maintainer call (e.g. build in the exact CI container, or make the check tolerant of non-deterministic bytes rather than exact-diff).

— PR comment commander

tupe12334:

— PR comment commander

Issue #266, opened 2026-06-18
HTTP API has no Host/Origin validation — a browser (incl. DNS rebinding) can trigger agent runs against the default loopback daemon


tupe12334:
## Problem

The daemon serves both a browser web UI (`GET /`) and a state-changing JSON API from the same origin, and `build_app` (`src/routes/http.rs`) wires only two middleware layers — `fs_location` and `logger`. There is **no `Host`/`Origin` header validation** anywhere.

Binding to loopback (`127.0.0.1:5784`) is correctly treated as the safe default (and #253 hardens the *non-loopback* bind case). But loopback is **not** a security boundary against a web browser: any web page the operator visits can issue requests to `http://127.0.0.1:5784` / `http://localhost:5784`. Combined with **DNS rebinding** (an attacker domain whose A record is re-pointed to `127.0.0.1`), a malicious site can reach the daemon even when it is bound only to loopback.

Because every route is unauthenticated and same-origin checks are absent, a browser-borne attacker can drive the state-changing endpoints:

- `POST /api/v1/routines` / `POST /api/v1/routines/{id}/trigger` — a routine fires its agent through a login shell that inherits the operator's environment (`GH_TOKEN`, API keys; see `src/routines/command.rs`). This is attacker-chosen code execution with the operator's credentials.
- `PUT/PATCH/DELETE` routine and cron-job routes — tamper with or delete scheduled work.
- `POST /api/v1/shutdown` — kill the daemon.

Simple-request shapes (e.g. `POST` with `text/plain`/`application/x-www-form-urlencoded`, or `GET`) are not blocked by the browser's preflight, and there is no CORS layer or token, so the request reaches the handler and executes.

This vector is **distinct from #253**: #253 is about the operator overriding `MOADIM_BIND_ADDR` to a public interface; this is about a browser attacking the **default loopback** deployment via cross-origin / DNS-rebinding requests. #253 explicitly leaves API authentication as a separate follow-up.

## Motivation

- The threat exists in the default, recommended configuration — no misconfiguration required.
- The blast radius is severe: arbitrary agent runs with the operator's tokens, not just data exposure.
- DNS-rebinding against localhost daemons is a well-known, repeatedly-exploited class (local model servers, dev daemons, etc.); a `Host`-allowlist guard is the standard, cheap mitigation.

## Acceptance criteria

[...]

PR #1147, opened 2026-07-13
fix(security): add missing 127.0.0.1:<port> to loopback Host allowlist

tupe12334:
## Summary

`allowed_hosts()` (in `src/middlewares/host_validation.rs`) adds `localhost:<port>` and `[::1]:<port>` to the DNS-rebinding-guard allowlist (issue #266) whenever the daemon's bind address carries a port — but never added the equivalent `127.0.0.1:<port>` entry, even though the bare `127.0.0.1` (no port) was already in the list.

A browser hitting the UI via `http://127.0.0.1:<port>` instead of `http://localhost:<port>` sends `Host: 127.0.0.1:<port>`, which fell outside the allowlist and got rejected with `403 Forbidden` — while the functionally identical `localhost:<port>` request passed. No existing test exercised the port-suffixed loopback branch, so this went unnoticed.

## Changes

- `allowed_hosts()`: push `127.0.0.1:<port>` alongside the existing `localhost:<port>` / `[::1]:<port>` entries.
- Added `allowed_hosts_includes_port_suffixed_loopback_entries_when_bind_addr_has_port`, asserting all three port-suffixed loopback spellings are present.
- Extended `allowed_hosts_skips_port_suffixed_entries_when_bind_addr_has_no_port` to also assert `127.0.0.1:` is absent when there's no port, mirroring its existing `localhost:`/`[::1]:` assertions.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo test --bin moadim host_validation` (13 passed)
- [x] `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` (100% line coverage)

### DIFF
--- a/.changeset/fix-allowed-hosts-loopback-ip-port.md
+++ b/.changeset/fix-allowed-hosts-loopback-ip-port.md
@@ -1,0 +1,13 @@
+---
+"moadim": patch
+---
+
+fix(security): add missing `127.0.0.1:<port>` entry to the loopback `Host`/`Origin` allowlist
+
+`allowed_hosts()` added `localhost:<port>` and `[::1]:<port>` alongside the bare bind address
+when the daemon's bind address carries a port, but never added the equivalent
+`127.0.0.1:<port>` entry — even though the bare `127.0.0.1` (no port) was already allowed. A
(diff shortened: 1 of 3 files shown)


tupe12334:
LGTM

tupe12334:

PR #1165, opened 2026-07-13
chore(lint): enable clippy::use_self in the ui crate

tupe12334:
## Why

The root crate denies `clippy::use_self`, but the `ui` crate has its own `[lints.clippy]` table and doesn't inherit root's extended deny-list — so `use_self` never applied to `ui/src` despite CI's `clippy` job running `--workspace`. This follows the same pattern as the several prior "mirror this root lint into ui" PRs already merged (`doc_markdown`, `if_not_else`, `manual_let_else`, etc.).

## What

- Add `use_self = "deny"` to `ui/Cargo.toml`.
- Fix the 135 violations it surfaced (enum/type name repeated inside its own impl block, e.g. `RGroupBy::Status` → `Self::Status`) via `cargo clippy --fix --allow-dirty -p ui`, across `refresh.rs`, `routines/state.rs`, `routines/filter.rs`, `overview_attention.rs`, and `schedule_heatmap_grid.rs`.
- Regenerate `prebuilt.html` via `trunk` (same as CI's `prebuilt-html-fresh` check) since `ui/src` changed.
- Add a changeset (patch — no behavior change).

## Test plan

- [x] `cargo clippy --workspace --all-targets --all-features` clean
- [x] `cargo fmt --check` clean
- [x] `cargo test -p ui --all-features` — 287 passed
- [x] `cargo check` (regenerates `prebuilt.html` via trunk) — committed the fresh output


tupe12334:

— PR comment commander

Issue #335, opened 2026-06-19
No global cap on concurrent routine runs: a shared cron minute (e.g. */5) launches an unbounded thundering herd of agent sessions


tupe12334:
## Problem

Every routine fire spawns its own agent run in a detached `tmux` session (`tmux new-session -d ...` in `src/routines/command.rs`), and the trigger/handler path has **no global cap on how many runs may be alive at once**. There is no semaphore, queue, or max-concurrent-runs guard anywhere in `src/routines/` (grep for `semaphore` / `max_concurr` / `concurrent` / `running_count` returns nothing).

Because routines are driven by the OS crontab, fires naturally **align on the same minute boundary**. With the common defaults (`*/5 * * * *`, `0 * * * *`) and a handful of routines, many agent sessions launch *simultaneously* on the same tick — a thundering herd. Each session is a full agent process (Claude Code / Codex / Hermes) that can clone repos, hold a long context, and burn provider API rate limits. N routines on a shared schedule => N concurrent heavyweight processes with nothing bounding N.

Failure modes:
- CPU/RAM exhaustion on the host when many runs collide on one tick.
- Provider API rate-limit / quota bursts (all agents hit the API at once).
- The per-run max-runtime watchdog does not help — it bounds *duration*, not *fan-out*.

This is **distinct from #141** (per-routine overlap: the *same* routine re-firing before its previous run finishes). That guard prevents one routine from stacking on itself, but does nothing to bound the total number of *different* routines running concurrently. Even with #141 fully fixed, 20 distinct routines sharing `0 * * * *` still launch 20 agents at once.

## Motivation

A daemon meant to run unattended on a personal machine should degrade gracefully under load instead of forking an unbounded number of agent processes on a shared cron minute. A global concurrency limit (with queueing or graceful deferral) makes resource use predictable and protects shared API quotas.

## Acceptance criteria

- [ ] Introduce a configurable global cap on concurrently-running routine agent sessions (e.g. `MOADIM_MAX_CONCURRENT_RUNS`, with a sane default such as 4).
- [ ] When a fire would exceed the cap, the run is **queued/deferred or skipped with a logged reason** rather than launched — choose and document one policy (queue vs. skip-this-tick).
- [ ] The count of live runs is derived from actual session liveness (reuse the existing `tmux_session_alive` seam), not just an in-memory counter that can drift after a crash.
- [ ] Behavior is covered by tests using the existing tmux/process test seams (no real `tmux` server).
[...]

PR #1166, opened 2026-07-13
feat(routines): expose the global concurrency cap through the UI/REST

### LINKED ISSUE #1155: Expose MOADIM_MAX_CONCURRENT_RUNS (global routine concurrency cap) in the UI
## Context
#1154 made the global routine concurrency cap (`MOADIM_MAX_CONCURRENT_RUNS`, `src/routines/concurrency_cap.rs`) default to unlimited (`0`). Today the only way to set it is the env var — there's no way to view or change it from the UI/REST.

## Request
Let the cap be configured through the UI/REST and persisted in a **local, machine-specific file** (not git-tracked, not synced) — same tier as `~/.config/moadim/machine.local.toml`, not the git-trackable routines/agents config.

## Suggested approach
Follow the existing pattern in `src/machine/mod.rs` / `src/paths/mod.rs` (`machine_config_path()` → `machine.local.toml`, already gitignored via the `*.local.*` rule in `config_gitignore_path()`):
- A small `MaxConcurrentRunsToml { value: Option<usize> }`-shaped struct, read with `toml::from_str` and written with `toml::to_string_pretty` via `crate::utils::atomic::atomic_write` (crash-safe, matches every other settings writer).
- Precedence: `MOADIM_MAX_CONCURRENT_RUNS` env var (existing, for ops/CI) > local file value (new, for UI) > `DEFAULT_MAX_CONCURRENT_RUNS = 0` fallback — mirrors how `current_machine()` layers `MOADIM_MACHINE` over its own local file.
- REST: a GET/PUT pair mirroring `get_current_machine`/`put_machine` in `src/routes/http_settings_routes.rs`.
- Simplest storage: either a new tiny `settings.local.toml`, or just add the field to the existing `machine.local.toml` if a single extra scalar doesn't justify a new file (ponytail: skip a generic key-value settings store for one field — only add one if a second UI-configurable local setting shows up).

## Open questions
- Does changing it via the UI take effect immediately (next trigger check reads the file live, like the env var does today) or require a restart?
- New file vs. extending `machine.local.toml` — worth deciding based on whether more per-machine UI settings are expected soon.

tupe12334:
## Summary

`MOADIM_MAX_CONCURRENT_RUNS` (the global routine concurrency cap, #335) was previously only
configurable via the environment variable — no way to view or change it from the UI/REST, per
#1155.

Following the exact pattern the issue points at (`src/machine/mod.rs` / `machine.local.toml` /
`get_current_machine`/`put_machine` in `http_settings_routes.rs`):

- `machine.local.toml` gets a second field, `max_concurrent_runs: Option<usize>` (a UI/REST
  override, gitignored, machine-local — same tier as the existing machine name).
- `set_machine`/the new `set_max_concurrent_runs_override` now read-modify-write the shared toml
  file instead of overwriting it wholesale, so setting one field never clobbers the other (this
  was a latent bug in `set_machine` before this PR — harmless until a second field existed).
- `GET`/`PUT /config/max-concurrent-runs` REST endpoints, wired the same way as
  `/config/user-prompt`.
- A "GLOBAL CONCURRENCY CAP" card on the Settings page (number input, empty = unbounded).

**Precedence** (per the issue's suggested approach): `MOADIM_MAX_CONCURRENT_RUNS` env var
(ops/CI) > the persisted UI/REST override > unbounded (`0`). Takes effect on the next trigger
check — `max_concurrent_runs()` reads live from disk/env, no restart required (answers the
issue's first open question).

Kept the storage as one extra field on `machine.local.toml` rather than a new file, per the
issue's own ponytail note — no second UI-configurable local setting exists yet to justify a
generic settings store.

Not included: a `#1156`-style separate "resolved-from" indicator distinguishing env vs. file vs.
default in the UI — the response does expose both `value` (fully resolved) and `override_value`
(the raw persisted override) so a future UI affordance can build on it without an API change.

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo fmt --check`
- [x] `cargo test --workspace` (1042 + 293 passed) — new tests cover: file-override precedence in
      `concurrency_cap.rs`, read/write/preservation round-trips in `machine/mod.rs` (including
      the read-modify-write fix for both fields), the new REST route (get/put/round-trip/clear/500
      on write failure), and the UI's pure `parse_cap` helper.
- [x] `linecheck --max-lines 500 $(find src ui/src -name '*.rs')`
- [x] `pnpm --filter client typecheck && lint && test` (regenerated `schema.gen.ts` from the
      updated `apis/openapi.json` — typechecks clean, 312 client tests pass)
- [x] Full `.githooks/pre-push` end to end (coverage, linecheck, client checks, changeset check)

Closes #1155


tupe12334:

— PR comment commander

PR #1169, opened 2026-07-13
test(ui): cover day_timeline's fire_times cron logic

tupe12334:
## Summary
- `fire_times` in `ui/src/day_timeline.rs` is the pure cron-to-fire-times logic behind the routines day timeline: it handles a midnight-boundary seed trick (stepping back one second so a fire at exactly `00:00:00` isn't dropped), filters out spillover from adjacent days, and caps iteration at `MAX_FIRES` for pathological schedules.
- Every other pure-logic file in `ui/src` (e.g. `schedule_heatmap.rs`, `cron_utils.rs`, `command_palette_match.rs`) has a companion `#[cfg(test)] #[path = "..."]` test module. `day_timeline.rs` had none — this subtle date logic could silently break in a refactor with nothing to catch it.
- Adds `ui/src/day_timeline_tests.rs` covering: multiple fires in a day (ordering), the midnight-exactly-at-boundary case, adjacent-day exclusion, unparseable schedules, and the `MAX_FIRES` cap.
- No production code behavior change — `day_timeline.rs` only gains the `mod` wiring at the bottom, matching the existing convention.

## Test plan
- [x] `cargo test -p ui day_timeline` — 5 new tests pass
- [x] `cargo clippy -p ui --all-targets --all-features` — clean
- [x] `cargo fmt --all -- --check` — clean

### DIFF
--- a/.changeset/day-timeline-fire-times-coverage.md
+++ b/.changeset/day-timeline-fire-times-coverage.md
@@ -1,0 +1,5 @@
+---
+"moadim": patch
+---
+
+Add host-side unit tests for the day timeline's `fire_times` (`ui/src/day_timeline.rs`), covering multi-fire schedules, the midnight-boundary seed, adjacent-day filtering, unparseable schedules, and the `MAX_FIRES` cap. This logic previously had no test module, unlike every other pure-logic file in the `ui` crate. No behavior change.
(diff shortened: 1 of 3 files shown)


tupe12334:

— PR comment commander

PR #1172, opened 2026-07-14
chore(lint): enable clippy::cast_lossless in the workspace

tupe12334:
## Summary
- Adds `cast_lossless = "deny"` to both the root crate's and the `ui` crate's `[lints.clippy]` tables, rejecting `as` casts that widen without loss (e.g. `u32 as i64`) in favour of `From`/`Into`.
- An `as` cast stays silently legal — and silently starts truncating — if the source or target type ever changes size; `i64::from(x)` is the same widening but fails to compile the moment it would no longer be lossless.
- Fixes the single violation this surfaced workspace-wide, in `ui/src/routines/calendar.rs`'s week-grid start calculation: `first.weekday().num_days_from_sunday() as i64` → `i64::from(first.weekday().num_days_from_sunday())`. No behavior change.
- Follows the repo's established one-lint-per-PR convention (see #1165, #1140, #1138 and ~29 already-merged commits of this shape).
- `prebuilt.html` regenerated via `trunk 0.21.14` to stay in lockstep with the `ui/` source change (per `prebuilt-ui.yml`).

## Test plan
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo test --workspace` (1035+287 passed)
- [x] `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` (100% line coverage)
- [x] `linecheck --max-lines 500 $(find src ui/src -name '*.rs')`
- [x] `typos` on changed files
- [x] `cargo check` to confirm `prebuilt.html` rebuilds byte-identical to the committed copy


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails, but root cause is a CI workflow bug, not a staleness issue in this PR's diff.

`.github/workflows/prebuilt-ui.yml` never installs pnpm/Node before running `cargo check`. The build log shows `warning: moadim@1.2.0: pnpm not found; React client not built`. Its sibling check `prebuilt-client.yml` (and `publish.yml`, at release time) both install pnpm via `pnpm/action-setup` + `actions/setup-node` before building — `prebuilt-ui.yml` is missing those same two steps. Without pnpm, the client falls back to whatever's already committed, so a locally-built `prebuilt.html` (with pnpm present) diverges from what CI produces — this affects every PR touching workspace-wide files, regardless of the actual diff.

— PR comment commander

PR #1195, opened 2026-07-14
chore(lint): enable clippy::or_fun_call in the ui crate

tupe12334:
## Summary

Adds `or_fun_call = "deny"` to the `ui` crate's `[lints.clippy]` table, mirroring the same lint already enabled workspace-root-side in `Cargo.toml`. The `ui` crate has its own `[lints.clippy]` table (no `workspace = true` inheritance), so this lint never applied to `ui/src` despite CI's `clippy` job running `--workspace`.

`or_fun_call` rejects a function call passed directly as the fallback argument to `unwrap_or`/`ok_or`/`and`/`or`-style methods (e.g. `opt.unwrap_or(expensive())`) in favor of the lazy `_else` form (`opt.unwrap_or_else(expensive)`) — the eager form always evaluates the fallback even on the common path where the value is already present.

The `ui` crate is already clean under this lint (zero violations surfaced), so this is a pure lockstep/config change — no behavior change, no source edits beyond `ui/Cargo.toml`.

This follows the same pattern as the already-merged/open root-crate → ui-crate lint lockstep PRs (e.g. #1165 `use_self`, #1140 `derive_partial_eq_without_eq`, #1138 `needless_pass_by_value`), for a lint not covered by any of those.

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean (0 new violations)
- [x] `cargo test --workspace` — 1054 + 299 tests pass
- [x] Added `.changeset/enable-clippy-or-fun-call-ui.md` (patch, `moadim` package)

### DIFF
--- a/.changeset/enable-clippy-or-fun-call-ui.md
+++ b/.changeset/enable-clippy-or-fun-call-ui.md
@@ -1,0 +1,10 @@
+---
+"moadim": patch
+---
+
+chore(lint): enable clippy::or_fun_call in the ui crate
+
+Mirrors the root crate's `or_fun_call = "deny"` (see `Cargo.toml`). The `ui` crate has its own
+`[lints.clippy]` table and doesn't inherit root's extended deny-list, so this never applied to
+`ui/src` despite CI's `clippy` job running `--workspace`. The `ui` crate is already clean under
+it (zero violations), so `deny` locks that in. No behavior change.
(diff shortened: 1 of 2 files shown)


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails — same root cause already diagnosed on #1165/#1143/#1142/#1140/#1138: `.github/workflows/prebuilt-ui.yml` never installs pnpm/Node before running `cargo check`, so the client falls back to whatever's already committed and diverges from a locally-regenerated `prebuilt.html`. Not caused by this PR's diff (pure Rust lint change). Fix is a workflow-file change with repo-wide blast radius — left for a maintainer to apply rather than automated here.

— PR comment commander

PR #1209, opened 2026-07-15
chore(lint): enable clippy::missing_docs in the ui crate

tupe12334:
## Why

Root's `Cargo.toml` denies `missing_docs` in its `[lints.rust]` table (requiring a doc comment on every public item), and CI's `clippy`/`doc` jobs run `--workspace`. But `ui` is a member crate with its own `[lints]` table — Cargo doesn't propagate root lint config to member crates without `workspace = true` inheritance — so `missing_docs` silently never applied to `ui/src` despite looking covered in CI. This is the same gap this repo has closed one lint at a time for `ui` before (e.g. #1206, #1172, #1137): a lint enabled root-side that quietly never reached the UI crate.

## What changed

- Enabled `missing_docs = "deny"` in `ui/Cargo.toml`'s new `[lints.rust]` table.
- Added the 34 missing doc comments this surfaced in `ui/src/main.rs`: the crate-level doc, the `Route` enum + variants, `ShellState` + its fields, `ShellAction` + its variants/fields, and the `App`/`Nav` function components.
- Split `ShellState`/`ShellAction`/their `Reducible` impl out into a new `ui/src/shell_state.rs` module — adding the required doc comments pushed `main.rs` past the workspace's 500-line-per-file `linecheck` gate (`.github/workflows/lint.yml`), so this move keeps it compliant without touching any other file's line count.
- `prebuilt.html` was regenerated by the `ui` crate's build script to reflect the `ui/src` changes (matches this repo's convention of committing it alongside `ui/`-touching PRs).

No behavior change — doc comments and a straight module split only.

## Test plan

- [x] `cargo fmt --check`
- [x] `linecheck --max-lines 500` over `src` + `ui/src`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items`
- [x] `cargo build --workspace`
- [x] `cargo test --workspace` (1054 + 307 passed)


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff (ui/ doc-comment additions for `clippy::missing_docs`). Rebuilding locally with the pinned toolchain (trunk 0.21.14, rustc stable) reproduces byte-for-byte the committed `prebuilt.html` — no diff — so the mismatch only shows up in CI's Linux runner, not on this (macOS) rebuild environment. Likely wasm-bindgen/trunk output isn't byte-reproducible across host platforms. Getting CI's exact bytes would require rebuilding inside a matching Linux environment or a temporary CI change to inspect the artifact, which is outside the automated CI-fix scope (changeset/behind-main only). Needs a human rebuild on Linux (or in CI) and commit of the result.

— PR comment commander

PR #1212, opened 2026-07-15
chore(lint): enable clippy::ref_option workspace-wide

tupe12334:
## Why

Continuing this repo's established pattern (see #1200, #1206, #1202, #1203, and others) of
incrementally locking in one more `clippy::pedantic` lint at a time, in both `Cargo.toml`
(root crate) and `ui/Cargo.toml` (the `ui` crate has its own `[lints.clippy]` table and does
not inherit the root's extended deny-list, so a root-side `deny` alone never applies to
`ui/src` despite CI's `clippy` job running `--workspace`).

`clippy::ref_option` flags a `&Option<T>` function parameter in favor of the more idiomatic
`Option<&T>` — the latter lets a caller pass a borrow directly (`Some(&x)`) instead of first
having to own or clone into an `Option`, and is cheaper/clearer to call.

## What changed

- Enabled `ref_option = "deny"` in both `Cargo.toml` and `ui/Cargo.toml`.
- The root `moadim` crate was already clean — `deny` just locks that in, no code changes needed.
- The `ui` crate surfaced exactly one violation: `schedule_label` in
  `ui/src/command_palette_match.rs` took `human: &Option<String>` only to immediately match on
  it by reference. Changed the signature to `Option<&String>`, updated its one call site
  (`routine.schedule_description.as_ref()`), and updated the existing unit test in
  `ui/src/command_palette_tests.rs` to pass `Option<&String>` values. No behavior change.

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes clean
- [x] `cargo test -p ui` — all 307 tests pass, including the updated `schedule_label_prefers_human_then_raw_then_dash`
- [x] Added a `.changeset/enable-clippy-ref-option.md` patch changeset


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff — same class of issue diagnosed on #1209: rebuilding locally reproduces the committed `prebuilt.html` byte-for-byte, so the mismatch is CI-Linux-runner-specific (non-reproducible wasm-bindgen/trunk output across host platforms), not something this PR's diff broke. Needs a human rebuild on Linux (or in CI) and commit of the result — outside the automated CI-fix scope (changeset/behind-main only).

— PR comment commander

tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff — same class of issue diagnosed on #1209/#1214/#1216/#1219: rebuilding locally reproduces the committed `prebuilt.html` byte-for-byte, so the mismatch is CI-Linux-runner-specific (non-reproducible wasm-bindgen/trunk output across host platforms), not something this PR's diff broke. Merge conflict against main has been resolved (normal merge commit). Needs a human rebuild on Linux (or in CI) and commit of the result — outside the automated CI-fix scope (changeset/behind-main only).

— PR comment commander

PR #1214, opened 2026-07-15
fix(ui): regenerate stale prebuilt.html to match ui/src

tupe12334:
## What

Regenerates \`prebuilt.html\` (the trunk-built, self-contained fallback UI bundle embedded by \`build.rs\`) so it matches the current \`ui/src\`.

## Why

\`prebuilt.html\` was last regenerated at 5e63928 (#1172), but \`ui/src\` has changed since (e.g. #1143, the day-detail popover) without a matching rebuild. The \`prebuilt-html-fresh\` CI check (\`.github/workflows/prebuilt-ui.yml\`) only runs on PRs that touch \`ui/**\`, so this drift on \`main\` itself went unnoticed — and now every PR that touches \`ui/\` fails that check regardless of its own content, including #1212 and #1209 currently open.

Regenerated locally with the CI-pinned \`trunk 0.21.14\` (\`cargo check\`, same as the workflow) and committing the resulting diff — no changes to \`ui/src\` or any other source.

No changeset needed: \`.changeset\`/the changelog gate only fires on \`src/**\`, \`ui/**\`, \`client/**\`, and this PR touches none of those — it's a generated-artifact-only fix.

---


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff — same class of issue diagnosed on #1209/#1212: rebuilding locally reproduces the committed `prebuilt.html` byte-for-byte, so the mismatch is CI-Linux-runner-specific (non-reproducible wasm-bindgen/trunk output across host platforms), not something this PR's diff broke. Needs a human rebuild on Linux (or in CI) and commit of the result — outside the automated CI-fix scope (changeset/behind-main only).

— PR comment commander

PR #1215, opened 2026-07-15
fix: refuse unauthenticated remote bind without explicit opt-in

### LINKED ISSUE #253: Unauthenticated HTTP API can be bound to any interface via MOADIM_BIND_ADDR with no guard or warning — remote callers can schedule/trigger agent runs carrying the operator's tokens
## Problem

The daemon's HTTP API is **unauthenticated** — `build_app` (`src/routes/http.rs`) wires only two middleware layers, `fs_location` and `logger`; there is no auth/token check anywhere in `src/`. Every route is open, including state-changing ones: `POST /routines`, `PUT/PATCH/DELETE /routines/{id}`, `POST /routines/{id}/trigger`, the cron-job equivalents, and `POST /shutdown`.

That is a fine default **because the server binds to loopback** (`BIND_ADDR = "127.0.0.1:5784"`, `src/cli.rs`). But the bind address is overridable with zero validation:

```rust
// src/cli.rs
pub fn bind_addr() -> String {
    std::env::var(BIND_ADDR_ENV).unwrap_or_else(|_| BIND_ADDR.to_string())
}
```

`run_server` (`src/main.rs`) hands that string straight to `TcpListener::bind(...)`. So `MOADIM_BIND_ADDR=0.0.0.0:5784` silently exposes the **entire unauthenticated API on every network interface**, with no warning at startup and no opt-in gate.

The blast radius is not just data: a routine's prompt and repositories are arbitrary, and a routine fires its agent through a login shell that inherits the operator's environment — `GH_TOKEN`, API keys, etc. (documented in `src/sync/routines.rs` / `src/routines/command.rs`). So anyone who can reach a non-loopback-bound daemon can `POST /routines` (or `/trigger`) to make an AI agent run an attacker-chosen task **with the user's tokens and shell credentials**, or `POST /shutdown` to kill the daemon. That is effectively remote code execution / credential exfiltration.

## Motivation

- The safe default (loopback) is one undocumented env var away from a fully-exposed, unauthenticated control plane — and the failure is silent: the server starts normally and logs nothing alarming.
- The API can schedule and trigger arbitrary agent runs carrying the operator's secrets, so the consequence of accidental exposure is severe, not cosmetic.
- This is distinct from #171 (which only fixed *reporting* the overridden address) — nothing today guards or even warns about *what interface* is being exposed.

## Acceptance criteria

- [ ] At startup, detect when the resolved bind address is **non-loopback** (parse the host with `std::net::IpAddr` and check `!is_loopback()`; treat `0.0.0.0` / `::` as non-loopback) and emit a prominent `log::warn!`/`error!` line stating the API is unauthenticated and now reachable off-host.
- [ ] Gate non-loopback binds behind an explicit opt-in (e.g. require an env var like `MOADIM_ALLOW_REMOTE=1`, or refuse to start otherwise) so exposure is never accidental — decide and document the exact behavior (warn-only vs. refuse).
- [ ] Document the risk and the opt-in next to `MOADIM_BIND_ADDR` in the README, noting the API has no authentication and should stay on loopback unless deliberately fronted by an authenticating proxy.
- [ ] Tests cover the address classification: loopback (`127.0.0.1`, `::1`) → allowed/no warning; non-loopback (`0.0.0.0`, a LAN IP) → warning/refusal per the chosen behavior.
- [ ] (Optional, larger follow-up) Track adding real authentication (shared token header) as a separate issue if remote use is intended to be a supported mode rather than a foot-gun.

---
*This issue was filed as a suggestion by the "Suggest new GitHub issues (daemon + landing)" routine of moadim. It is intentionally left UNASSIGNED — approve it by self-assigning. cc @tupe12334*

tupe12334:
## Why

The REST/MCP API has no authentication (#504). `#1039` added a startup `log::warn!` when `MOADIM_BIND_ADDR` resolves to a non-loopback address, but binding was never gated — `MOADIM_BIND_ADDR=0.0.0.0:5784` would still start and silently expose unauthenticated routine create/edit/delete/trigger (and `/shutdown`) to the network. Since a triggered routine launches an agent through a login shell carrying the operator's environment (`GH_TOKEN`, API keys, …), this is effectively RCE / credential exfiltration for anyone who can reach the bound address.

## What

- `cli::classify_bind(addr: &str, allow_remote: bool) -> BindDecision` — a pure decision function: `Loopback` (start normally), `RemoteAllowed` (non-loopback + opted in, start with a warning), `RemoteRefused` (non-loopback, not opted in, refuse to start). Built on the existing `bind_addr_is_loopback` from #1039.
- `cli::remote_bind_allowed()` reads the new `MOADIM_ALLOW_REMOTE` env var; it must be exactly `"1"` — any other value (unset, `true`, `yes`, …) is treated as not opted in, so a typo fails closed.
- `run_server()` in `src/main.rs` now matches on `classify_bind` at startup: refuses to start (`anyhow::bail!` with an explanatory error) on `RemoteRefused`, or logs a prominent `log::warn!` and continues on `RemoteAllowed`. Loopback binds are unaffected.
- README: documents `MOADIM_ALLOW_REMOTE` and the refuse-by-default behavior right next to the existing `MOADIM_BIND_ADDR` / bind-address security warning.
- CHANGELOG: `[Unreleased]` entry.
- Unit tests in `src/cli/tests.rs` cover `classify_bind` for loopback (v4/v6, both opt-in states), non-loopback without opt-in (refused), non-loopback with opt-in (allowed), and `remote_bind_allowed`'s exact-`"1"` matching (unset/`true`/`yes`/`0`/empty all rejected).

This chooses "refuse unless opted in" per the issue's stated preference, since this is RCE-adjacent. Auth for the API itself remains out of scope, tracked separately (#504).

## Testing

- `cargo build -p moadim` — clean
- `cargo test -p moadim` — 1060 passed, 0 failed (7 new: `classify_bind_*`, `remote_bind_allowed_*`)
- `cargo clippy -p moadim --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean

Closes #253


tupe12334:
Added the missing changeset (`unreleased-entry`/Changelog check was failing). One remaining CI failure is a genuine code issue from this PR's own diff, not automatable:
- `linecheck`: `src/cli/mod.rs` is now 521 lines and `src/cli/tests.rs` is 510 lines (threshold 500) — needs splitting, e.g. factoring the new `classify_bind`/`remote_bind_allowed` logic (and its tests) into their own module.

Needs a judgment call on how to split — beyond automated CI-fix scope.

— PR comment commander

PR #1216, opened 2026-07-16
fix(ui): flag snoozed routines in the routines calendar's day-detail popover

tupe12334:
## Why

The routines calendar's month grid already dims a routine's chip (amber, reduced opacity) when its next fire is suppressed by `snoozed_until` or a pending `skip_runs` counter — because that scheduled time will be *silently skipped* by `svc_trigger_scheduled` rather than actually running an agent. The server-side `.ics` feed (`src/routines/ical.rs`) applies the same logic and explicitly omits those fires so subscribers are never told about a run "that will silently no-op".

The day-detail popover (added in #1143 — click a day number to see every fire time that day) never got this treatment: it lists all of an enabled routine's fire times for the clicked day with zero indication that some of them belong to a currently-snoozed routine. A user could open the popover, see "09:00 — Nightly Cleanup", and reasonably expect it to run — when it's actually going to be skipped. This is a small but real regression in information carried over from the calendar cell into the newer popover.

## What

- `day_fire_rows` (`ui/src/routines/calendar.rs`) now takes a `now: DateTime<Local>` and returns an extra `snoozed: bool` per row, computed with the same `is_routine_snoozed` helper the month grid's chips already use — no new logic, just reusing the existing signal.
- The popover renders a `SNOOZED` badge on those rows (same `health-badge snoozed` class/wording used in the routines table) and dims the row, mirroring `.cal-chip.snoozed`'s existing amber/opacity treatment.
- `ui/styles.css`: two small rules for `.day-fire-row.snoozed`.
- Tests: updated the existing `day_fire_rows` tests for the new signature/tuple shape, and added three new tests covering snoozed-by-deadline, snoozed-by-skip-runs, and "no longer snoozed since the deadline passed".
- `prebuilt.html` regenerated via the pinned `trunk` 0.21.14 (matching `prebuilt-ui.yml`) to stay in lockstep with the `ui/src` change.

Frontend-only — no backend, API, or data-model changes; the manual "▶ RUN" button still works as before (it always was, and still is, a real manual trigger, independent of the scheduled-fire skip).

## Checks run locally

- `cargo fmt --all -- --check` — clean
- `cd ui && cargo clippy --all-targets --target wasm32-unknown-unknown -- -D warnings` — clean
- `cargo test --workspace` — 1055 (root) + 310 (ui, +4 net new) passed
- Changeset added: `.changeset/day-detail-popover-flag-snoozed-rows.md` (`patch`)

**Known caveat:** as with prior `ui/`-touching PRs (#1140/#1142/#1143), a locally-rebuilt `prebuilt.html` (macOS/aarch64 `trunk`) isn't guaranteed byte-identical to what CI's `ubuntu-latest` `trunk`/`wasm-bindgen` toolchain produces from the same source, so `prebuilt-html-fresh` may still flag a diff even though the source-level change is complete and correct — a pre-existing, platform-dependent CI quirk unrelated to this diff (see #1214).


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff — same class of issue diagnosed on #1209/#1212/#1214: rebuilding locally (pinned trunk 0.21.14) reproduces the committed `prebuilt.html` byte-for-byte, so the mismatch is CI-Linux-runner-specific (non-reproducible wasm-bindgen/trunk output across host platforms), not something this PR's diff broke. Needs a human rebuild on Linux (or in CI) and commit of the result — outside the automated CI-fix scope (changeset/behind-main only).

— PR comment commander

PR #1221, opened 2026-07-16
fix(ui,client): dormant filter facet ignores blank-string machine entries

tupe12334:
## Summary

The Routines "Dormant" health badge and the Overview's dormant KPI (`routine_health` in the Yew UI, `routineHealth` in the React client) already treat a `machines` list holding only blank/whitespace entries (e.g. `[""]`) as "no real machine assigned" — there's an existing test for it (`health_enabled_blank_machine_entry_is_dormant`). The Status **filter facet** (`RoutineStatusFacet::Dormant` / `"dormant"` in `matchesFilter`) never got the same treatment: it only checked for a strictly empty `machines` list.

Net effect: a routine with `machines: [""]` shows a **DORMANT** badge on its row and is counted in the Overview's dormant KPI tile, but picking **Status = Dormant** in the filter bar does not surface that row — a visible contradiction between what's badged as dormant and what the matching filter finds.

## Why

This is a real UI bug a user would hit directly: click the dormant KPI tile (which cross-filters to `Status: Dormant`) and a routine you can plainly see is badged DORMANT in the table is missing from the filtered view. It exists identically in both frontends since `client/` is a direct, faithful port of `ui/`.

## Changes

- `ui/src/routines/filter.rs`: `RoutineStatusFacet::Dormant` now uses the same `machines.iter().all(|m| m.trim().is_empty())` predicate `routine_health` already uses, plus a doc-comment update.
- `client/src/pages/routines/filter.ts`: same fix in `matchesFilter`'s `"dormant"` case.
- Added a regression test to each (`status_dormant_matches_a_blank_machine_entry_like_routine_health_does` / `"status dormant matches a blank machine entry like routineHealth does"`) mirroring the existing health test.
- Regenerated `prebuilt.html` / `prebuilt-client.html` (required by CI's `prebuilt-ui`/`prebuilt-client` checks whenever `ui/**`/`client/**` changes).

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --workspace --bin ui` — 70/70 `routines::filter` tests pass, including the new one
- [x] `pnpm --filter client typecheck`
- [x] `pnpm --filter client lint`
- [x] `pnpm --filter client test` — 318/318 tests pass, including the new one
- [x] `cargo build` / `pnpm --filter client build` to regenerate `prebuilt.html` / `prebuilt-client.html`


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff — same class of issue diagnosed on #1209/#1212/#1214/#1216/#1219/#1222: rebuilding locally (pinned trunk 0.21.14) reproduces the committed `prebuilt.html` byte-for-byte, so the mismatch is CI-Linux-runner-specific (non-reproducible wasm-bindgen/trunk output across host platforms), not something this PR's diff broke. Needs a human rebuild on Linux (or in CI) and commit of the result — outside the automated CI-fix scope (changeset/behind-main only).

— PR comment commander

PR #1222, opened 2026-07-16
chore(lint): enable clippy::redundant_clone workspace-wide

tupe12334:
## Why

`cargo clippy --workspace --all-targets -- -W clippy::redundant_clone -D warnings` surfaced 31
violations: a `.clone()` call whose result is immediately moved with no other use of the original
value still pending. Every case follows one of two shapes:

- An intermediate `let x = x.clone();` shadow, built so a `move` closure/`html!` prop could
  capture an owned copy — except `x` (the outer binding) turned out not to be used again
  afterward, so the clone was pure overhead over just moving `x` directly.
- A `field.clone()` (e.g. `created.routine.id.clone()`, `routines.clone()`) read straight into a
  `let` or passed as a constructor argument that never touches the original again.

Neither pattern changes behavior — Rust already lets you move a value into a closure directly when
nothing after it needs the original — so this is a pure cleanup: less code, one fewer `Rc` refcount
bump per fired `Callback`/`UseStateHandle` update in the affected `ui/` components, one fewer
`String`/`Vec` allocation per affected test helper call.

## What changed

- `Cargo.toml`: enable `clippy::redundant_clone = "deny"` (matches this repo's established pattern
  of enabling one previously-off lint per PR, documented inline next to the others already in the
  `[lints.clippy]` table).
- All 31 flagged call sites, split evenly between `ui/src/**` (14, e.g. `command_palette.rs`,
  `main.rs`, `settings.rs`) and root-crate test files (17, e.g. `mcp_tests.rs`,
  `service_flag_tests.rs`): removed the redundant `.clone()`, letting the existing `move` closure
  or constructor call take ownership of the original value instead.
- `prebuilt.html` regenerated via `trunk` (pinned v0.21.14, matching `prebuilt-ui.yml`) to match
  the recompiled `ui/` wasm output, per this repo's `prebuilt-html-fresh` CI gate.
- `.changeset/enable-clippy-redundant-clone.md` added (no behavior change, `patch`).

## Verification

All green locally, matching `.githooks/pre-push`:
- `cargo fmt --check`
- `cargo test --workspace` (1055 + 307 passed)
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` (100% line floor held)
- `linecheck --max-lines 500` over `src` + `ui/src`
- `pnpm --filter client typecheck / lint / test`


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff — same class of issue diagnosed on #1209/#1212/#1214/#1216/#1219: rebuilding locally (pinned trunk 0.21.14) reproduces the committed `prebuilt.html` byte-for-byte, so the mismatch is CI-Linux-runner-specific (non-reproducible wasm-bindgen/trunk output across host platforms), not something this PR's diff broke. Needs a human rebuild on Linux (or in CI) and commit of the result — outside the automated CI-fix scope (changeset/behind-main only).

— PR comment commander

PR #1223, opened 2026-07-16
fix(ui,client): omit blank machine entries from the Machine filter dropdown

tupe12334:
## Summary

`distinct_machines_r` (`ui/src/routines/filter_distinct.rs`) and `distinctMachines`
(`client/src/pages/routines/filter.ts`) build the option list for the Routines page's Machine
filter dropdown by collecting every raw string in each routine's `machines` list. Neither
filters out blank/whitespace-only entries.

The API already rejects blank `machines` entries on create/update (`validate_machines`, #600),
but routines written before that guard existed can still carry one (e.g. `machines: [""]`), and
`routineHealth` / `routine_health` already special-case exactly this shape as "no real machine
assigned" (dormant) — there are existing tests for that. The dropdown-option builders never got
the same treatment, so such a routine can add a stray, unlabeled blank `<option>` to the Machine
facet — distinct from the existing "Any" and "Unassigned" options, and meaningless if selected.

This is a different code path from #1221 (which fixes the *matching* logic for the Dormant
status facet); this PR fixes the *option-list* builders that populate the Machine dropdown
itself, so the two changes don't overlap.

## Changes

- `ui/src/routines/filter_distinct.rs`: `distinct_machines_r` now skips machine entries that are
  empty or whitespace-only.
- `client/src/pages/routines/filter.ts`: `distinctMachines` gets the same filter.
- Added a regression test to each (`distinct_machines_r_omits_blank_machine_entries` /
  `"distinct_machines omits blank machine entries"`).
- Regenerated `prebuilt.html` / `prebuilt-client.html` (required by CI whenever `ui/**`/`client/**`
  changes).

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --workspace --bin ui` — 308/308 `ui` tests pass, including the new one
- [x] `pnpm --filter client typecheck`
- [x] `pnpm --filter client lint`
- [x] `pnpm --filter client test` — 319/319 tests pass, including the new one
- [x] `cargo build` / `pnpm --filter client build` to regenerate `prebuilt.html` / `prebuilt-client.html`


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff — same class of issue diagnosed on #1209/#1212/#1214/#1216/#1219/#1221/#1222: rebuilding locally (pinned trunk 0.21.14) reproduces the committed `prebuilt.html` byte-for-byte, so the mismatch is CI-Linux-runner-specific (non-reproducible wasm-bindgen/trunk output across host platforms), not something this PR's diff broke. Needs a human rebuild on Linux (or in CI) and commit of the result — outside the automated CI-fix scope (changeset/behind-main only).

— PR comment commander

PR #1227, opened 2026-07-16
fix(client): restore abstime hover tooltips lost in the Yew-to-React port

tupe12334:
## Why

`client/` (the React dashboard) is being built out as a direct, file-by-file port of the original Yew/WASM UI (`ui/`) — most files carry an explicit "Direct port of `ui/src/...`" doc comment, and their behavior is expected to match.

`ui/src/cron_utils.rs`'s `abstime()` renders an absolute wall-clock timestamp (`"Jun 21, 2026 12:00"`) and is used everywhere the UI shows a relative "N ago"/"Nm ago" time, as the `title` attribute for a hover tooltip:

- `overview_recent_runs.rs` — STARTED cell
- `routines/row.rs` — UPDATED cell
- `routines/history.rs` — run-row title (`"{workbench} · {abstime}"`)
- `routines/sparkline.rs` — per-tick tooltip (`"{status} · {reltime} ({abstime})"`)

The TypeScript port never gained an `abstime` function at all, so the equivalent React components silently regressed:

- `RecentRunsTable.tsx` — STARTED cell has no tooltip
- `RoutineRow.tsx` — UPDATED cell has no tooltip
- `RoutineHistory.tsx` — run-row title carries only the workbench name, missing the absolute time
- `RunHistorySparkline.tsx` — per-tick tooltip is missing the `(abstime)` suffix

An operator hovering a relative time in the React client currently can't see the exact timestamp the Yew UI has always shown there.

## What

- Add `abstime()` to `client/src/lib/cronUtils.ts`, mirroring the Rust formatting (`"MMM DD, YYYY HH:MM"`) and its `0`/out-of-range fallbacks to `"—"`.
- Wire it into the four call sites above so their tooltips match the Rust UI again.
- Add unit tests for `abstime` mirroring `ui/src/cron_utils_tests.rs`'s cases (zero placeholder, a known instant, an out-of-range instant).

## Test plan

- [x] `pnpm --filter client typecheck` — passes
- [x] `pnpm --filter client lint` — passes
- [x] `pnpm --filter client test` — 320/320 passing (incl. 3 new `abstime` tests)
- [x] `cargo fmt --check` — clean (no Rust files touched)
- [x] Changeset added (`.changeset/client-abstime-tooltips.md`, patch)

### DIFF
--- a/.changeset/client-abstime-tooltips.md
+++ b/.changeset/client-abstime-tooltips.md
@@ -1,0 +1,13 @@
+---
+"moadim": patch
+---
+
+Fix the React client (`client/`) silently dropping the absolute-timestamp hover tooltip that the
+original Yew UI (`ui/`) shows next to every relative "N ago" time. `ui/src/cron_utils.rs`'s
+`abstime` had no TypeScript port at all, so the "STARTED"/"UPDATED" cells in
+`RecentRunsTable.tsx`/`RoutineRow.tsx` rendered no `title`, `RoutineHistory.tsx`'s run-row title
+carried only the workbench name, and `RunHistorySparkline.tsx`'s per-tick tooltip omitted the
+absolute time — all despite each file being documented as a "direct port" of its Rust
+counterpart. Adds `abstime` to `client/src/lib/cronUtils.ts` (mirroring the Rust formatting and
+its zero/out-of-range fallbacks) and wires it into the four call sites so hovering a relative time
+in the React client shows the same wall-clock timestamp the Yew UI has always shown.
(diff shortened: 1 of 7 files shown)


tupe12334:
Blocked: CI check `prebuilt-client-html-fresh` fails against this PR's own diff — same class of issue diagnosed on #1209/#1212/#1214/#1216/#1219/#1221/#1223 for the `ui/` bundle, here on the `client/` bundle instead: rebuilding locally (`pnpm install && cargo check`) reproduces a diff too, but every changed line is a whole-script-tag minifier re-emit (different variable-name hashing), not a content change — consistent with esbuild/vite output not being byte-reproducible across Node versions (CI uses Node 22, local rebuild environment differs). Needs a human rebuild on a matching environment (or in CI) and commit of the result — outside the automated CI-fix scope (changeset/behind-main only).

— PR comment commander

PR #1219, opened 2026-07-16
chore(lint): enable clippy::many_single_char_names workspace-wide

tupe12334:
## Why

`clippy::many_single_char_names` rejects a scope with 4+ single-character bindings in play at once — past the first couple, a reader has to hold each letter's meaning in their head instead of reading a name. Sibling addition to the growing list of `deny`d pedantic lints already in `Cargo.toml` (most recently `similar_names`, #1218).

## What

Surfaced one violation: `ui/src/routines/filter_tests.rs`'s `is_active_detects_each_facet` test built six single-letter `RoutineFilter` fixtures in the same scope (`q`, `s`, `a`, `m`, `r`, `t` — one per facet under test). Renamed each to a descriptive name: `query_filter`, `status_filter`, `agent_filter`, `machine_filter`, `repo_filter`, `tag_filter`.

Added the lint to both `Cargo.toml` (root) and `ui/Cargo.toml` (the `ui` crate has its own `[lints.clippy]` table with no workspace inheritance, per the existing pattern documented throughout that file).

No behavior change — test-only rename plus two new `deny` lint entries.

## Checks

- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo clippy -p ui --target wasm32-unknown-unknown --all-targets -- -D warnings` — clean
- `cargo test --workspace` — 307 passed

---
This pull request was opened by the Daemon + Landing auto-improve & auto-merge lint/docs PRs routine of moadim. cc @tupe12334

PR #1231, opened 2026-07-16
chore(lint): enable clippy::format_collect in the ui crate

tupe12334:
## Why

Root's `Cargo.toml` denies `clippy::format_collect` in its `[lints.clippy]` table (reject `.map(|x| format!(...)).collect::<String>()` in favour of folding a `write!`/`writeln!` directly into a `String` buffer), and CI's `clippy` job runs `--workspace`. But `ui` is a member crate with its own `[lints]` table — Cargo doesn't propagate root lint config to member crates without `workspace = true` inheritance — so `format_collect` silently never applied to `ui/src` despite looking covered in CI. This is the same gap this repo has closed one lint at a time for `ui` before (e.g. #1206, #1172, #1137, #1219).

## What changed

- Enabled `format_collect = "deny"` in `ui/Cargo.toml`'s `[lints.clippy]` table.

No behavior change.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy -p ui --all-targets -- -D warnings`


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails deterministically against this PR's own diff (`ui/` clippy `format_collect` fix), but not for the reason the check message implies.

Rebuilt locally with the pinned toolchain (`cargo install --locked --version 0.21.14 trunk && cargo build` on this PR's exact commit) — the result is byte-identical to the already-committed `prebuilt.html`, not to what CI produces. CI's own diff shows the same hash (`e4e863f`) on both the original run and a rerun, so it isn't a one-off flake either — it's a reproducible mismatch between my build environment and CI's (macOS vs `ubuntu-latest`), most likely non-deterministic `wasm-bindgen`/`trunk` output rather than genuinely stale source.

This needs verification in a Linux environment matching CI exactly (or a look at whether the `ui/` build is fully reproducible across platforms) — beyond automated CI-fix scope.

— PR comment commander

PR #1233, opened 2026-07-16
fix(ui): regenerate stale prebuilt.html

tupe12334:
## Why

`.github/workflows/prebuilt-ui.yml` (`prebuilt-html-fresh` job) rebuilds the
Yew UI with the pinned `trunk 0.21.14` on every PR touching `ui/**` and fails
if the freshly built `prebuilt.html` doesn't byte-match what's committed.

The committed `prebuilt.html` on `main` is currently stale: a clean
`cargo check` (which invokes `build.rs` → `trunk build --release`) produces a
different bundle than what's checked in. This means the gate fails on *any*
PR that touches `ui/**` — including PRs that never change `ui/src` (e.g. a
`ui/Cargo.toml`-only lint change), since the workflow's path filter matches
`ui/**` broadly.

Confirmed reproducible: two independent clean `cargo check` runs against
current `main` produce identical (but different-from-committed) bytes for
`prebuilt.html`.

## What

Regenerates `prebuilt.html` from current `ui/src` via `trunk build --release`
(same tool/version CI pins) and commits the fresh bytes. No source change —
UI behavior is unaffected.

## Test plan

- [x] `cargo check` (regenerates `prebuilt.html`, no further diff after)
- [x] `cargo fmt --check`
- [x] `cargo test --workspace` — 312 passed
- [x] Reproduced the drift twice from a clean tree before fixing, to rule out one-off non-determinism


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff — same underlying issue diagnosed on #1231: rebuilding locally (`cargo install --locked --version 0.21.14 trunk && cargo build`) produces a `prebuilt.html` that's byte-identical to what's committed here, not to what CI's `ubuntu-latest` runner produces. This PR's stated purpose is regenerating `prebuilt.html`, so the committed file was already rebuilt once — yet CI still reports a diff, pointing to non-reproducible `trunk`/`wasm-bindgen` output (e.g. embedded build-path debuginfo) across environments rather than genuinely stale source.

This needs investigation into the build's reproducibility (verifying in an environment matching CI exactly, or trimming embedded paths) — beyond automated CI-fix scope.

— PR comment commander

PR #1234, opened 2026-07-16
chore(lint): enable clippy::unnested_or_patterns in the ui crate

tupe12334:
## Why

Root's `Cargo.toml` denies `clippy::unnested_or_patterns` in its `[lints.clippy]` table (reject an or-pattern repeated across multiple match arms/parameters in favour of merging into one arm with a single nested or-pattern), and CI's `clippy` job runs `--workspace`. But `ui` is a member crate with its own `[lints]` table — Cargo doesn't propagate root lint config to member crates without `workspace = true` inheritance — so `unnested_or_patterns` silently never applied to `ui/src` despite looking covered in CI. This is the same gap this repo has closed one lint at a time for `ui` before (e.g. #1219, #1231).

## What changed

- Enabled `unnested_or_patterns = "deny"` in `ui/Cargo.toml`'s `[lints.clippy]` table.

No behavior change.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy -p ui --all-targets -- -D warnings`


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff (`ui/` clippy `unnested_or_patterns` fix), but not for the reason the check message implies — same underlying issue diagnosed on #1231. Rebuilt locally with the pinned toolchain (`cargo install --locked --version 0.21.14 trunk && cargo build` on this PR's exact commit) and the result is byte-identical to the already-committed `prebuilt.html`, not to what CI's `ubuntu-latest` runner produces — pointing to non-reproducible `trunk`/`wasm-bindgen` output across build environments rather than genuinely stale source.

— PR comment commander

PR #1235, opened 2026-07-16
chore(lint): enable clippy::large_stack_arrays in the ui crate

tupe12334:
## Why

Root's `Cargo.toml` denies `clippy::large_stack_arrays` in its `[lints.clippy]` table (reject a local array literal over 512000 bytes — a value that size belongs on the heap, not the stack), and CI's `clippy` job runs `--workspace`. But `ui` is a member crate with its own `[lints]` table — Cargo doesn't propagate root lint config to member crates without `workspace = true` inheritance — so `large_stack_arrays` silently never applied to `ui/src` despite looking covered in CI. This is the same gap this repo has closed one lint at a time for `ui` before (e.g. #1219, #1231, #1234).

A daemon process runs long-lived worker threads with a fixed, comparatively small stack, so an oversized stack array is a latent stack-overflow risk that only surfaces under the right call depth — a heap allocation fails safely instead. The `ui` crate compiles to wasm32, which also has a fixed, comparatively small stack, so the same risk class applies there too.

## What changed

- Enabled `large_stack_arrays = "deny"` in `ui/Cargo.toml`'s `[lints.clippy]` table.

No behavior change.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test`


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff (`ui/` clippy `large_stack_arrays` fix), but not for the reason the check message implies — same underlying issue diagnosed on #1231. Rebuilt locally with the pinned toolchain (`cargo install --locked --version 0.21.14 trunk && cargo build` on this PR's exact commit) and the result is byte-identical to the already-committed `prebuilt.html`, not to what CI's `ubuntu-latest` runner produces — pointing to non-reproducible `trunk`/`wasm-bindgen` output across build environments rather than genuinely stale source.

— PR comment commander

PR #1237, opened 2026-07-16
chore(lint): enable clippy::similar_names in the ui crate

tupe12334:
## Why

`ui/Cargo.toml` maintains its own `[lints.clippy]` table (it doesn't inherit `workspace = true` from root), so every extended deny-list entry added to the root crate's `Cargo.toml` has to be mirrored there by hand or it silently never applies to `ui/src` — even though CI's `clippy` job runs with `--workspace`. This repo has been closing that gap one lint at a time (see `redundant_clone`, `ref_option`, `use_self`, `many_single_char_names`, etc. already mirrored into `ui/Cargo.toml`).

`clippy::similar_names` is one of the still-missing entries: it's `deny`d workspace-root-side (`Cargo.toml`) but was never added to `ui/Cargo.toml`, so it doesn't actually apply to `ui/src` despite the workspace-wide clippy CI job. The lint rejects a binding whose name is a near-miss of another binding already in scope (e.g. `txt` next to `text`) — a real bug-prevention lint, since a reader (or a future edit) can mix the two up.

I checked the currently-open PR queue (`gh pr list`) and confirmed no open PR already covers `similar_names`; the ones in flight cover `large_stack_arrays`, `unnested_or_patterns`, and `format_collect` instead.

## What

- Add `similar_names = "deny"` to `ui/Cargo.toml`'s `[lints.clippy]` table, with the same documentation style as its neighboring entries.
- Add a changeset (`patch`).

## Verification

- `cargo clippy -p ui --target wasm32-unknown-unknown --all-targets -- -D warnings` — 0 violations surfaced; the `ui` crate was already clean under this lint, so this is a lint-config-only change (no `ui/src` edits needed).
- `cargo build -p ui --target wasm32-unknown-unknown` — builds clean.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (1063 passed), `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main.rs'` (100% line coverage), `linecheck --max-lines 500 …` — all pass.
- `pnpm --filter client typecheck`, `pnpm --filter client lint`, `pnpm --filter client test` (322 passed) — all pass (unaffected by this change, run per the pre-push hook's full check list).

No behavior change — this only tightens lint coverage to match the root crate.


tupe12334:
Blocked: CI check `prebuilt-html-fresh` fails against this PR's own diff (`ui/` clippy `similar_names` fix), but not for the reason the check message implies — same underlying issue diagnosed on #1231. Rebuilt locally with the pinned toolchain (`trunk 0.21.14`, `cargo install --locked` version, on this PR's exact commit) and the result is byte-identical to the already-committed `prebuilt.html`, not to what CI's `ubuntu-latest` runner produces — pointing to non-reproducible `trunk`/`wasm-bindgen` output across build environments rather than genuinely stale source.

(`unreleased-entry` also failed on this PR but was a transient GitHub API error on the paths-filter step, not a real changeset issue — resolved by rerunning the job.)

— PR comment commander

PR #1242, opened 2026-07-17
fix(ui,client): Unassigned machine facet matches blank machine entries

tupe12334:
## Problem

The Machine filter's **Unassigned** option checks the raw `machines` array on both sides of the codebase:

- `ui/src/routines/filter.rs:243` — `RoutineMachineFacet::Unassigned if !r.machines.is_empty()`
- `client/src/pages/routines/filter.ts:156` — `f.machine.kind === "unassigned" && machines.length > 0`

A routine created before the `validate_machines` guard (#600) can still carry a blank/whitespace-only entry like `[""]`. That exact shape is already treated as "no real machine assigned" everywhere else that matters — the Dormant status facet (`filter.rs:224`, `filter.ts:140`), `routineHealth`, and the Machine filter dropdown (#1221, #1223) all use `.trim().is_empty()`/`.trim() === ""`. The Unassigned *facet* was the one place still using a raw emptiness/length check, so such a routine shows up as Dormant and is hidden from the dropdown, but selecting "Unassigned" in the Machine filter still wouldn't find it.

## Fix

Bring both sides in line with the established convention:

- Rust: `!r.machines.iter().all(|m| m.trim().is_empty())`
- TS: `!machines.every((m) => m.trim() === "")`

Added a blank-entry (`[""]`) case to the existing `machine_unassigned_matches_only_empty_machines` test on both sides.

## Testing

- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` — all pass
- `pnpm --filter client typecheck && pnpm --filter client lint && pnpm --filter client test` — all pass


tupe12334:
Blocked: CI check `prebuilt-html-fresh` still fails after regenerating both prebuilt bundles (pushed in 15b2dab) — `prebuilt-client-html-fresh` now passes (client/ was genuinely stale, fixed), but the `ui/` trunk/wasm-bindgen bundle rebuilt locally still doesn't match what CI's `ubuntu-latest` runner produces. Same non-reproducible cross-platform `trunk`/`wasm-bindgen` output issue diagnosed on #1231 and its siblings (#1233–#1239) — not something this PR's own diff caused.

— PR comment commander

PR #1251, opened 2026-07-17
fix(client): realign @vitejs/plugin-react, react-dom, @types/react-dom with their peer deps

tupe12334:
## Problem

Two independent, previously-merged automated dependency bumps left `client/`'s test toolchain with unsatisfiable peer dependencies, and stacked to fully break `pnpm --filter client test`:

- `@vitejs/plugin-react@6.0.3` (bumped in #1191) requires `vite@^8.0.0`, but `client/package.json` pins `vite@^6.0.5`. This crashed `vitest` before it could even load its config (`ERR_PACKAGE_PATH_NOT_EXPORTED` resolving `vite`'s `./internal` subpath, which only exists in vite 8).
- `react-dom@19.2.7` / `@types/react-dom@19.2.3` (bumped in #1187) require `react@^19` / `@types/react@^19`, but `client/package.json` pins the `18.3.x` line. Once the vite issue above is fixed, this crashes every test that actually renders a component (`react-dom-client.development.js` reads an internal field that doesn't exist when paired with React 18).

Together, `pnpm --filter client test` — part of both CI's `client-test` job (`.github/workflows/test.yml`) and step 7 of the local pre-push hook (`.githooks/pre-push`) — could not run at all, for any contributor, regardless of what they changed.

## Fix

Pin back to the last versions actually compatible with this repo's other pins:
- `@vitejs/plugin-react`: `^6.0.3` → `^5.2.0` (last major supporting `vite@^6`)
- `react-dom`: `^19.2.7` → `^18.3.1` (matches the `react@^18.3.1` already pinned)
- `@types/react-dom`: `^19.2.3` → `^18.3.7` (matches `@types/react@^18.3.17`)

No application code changes — this is purely a dependency-pin correction plus the resulting `pnpm-lock.yaml` update.

## Verification

- `pnpm --filter client test` — 329 passed (previously failed to even start)
- `pnpm --filter client typecheck` — clean
- `pnpm --filter client build` — clean
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — all clean (unaffected by this change, confirmed for completeness)

Note: `pnpm --filter client lint` still reports 8 pre-existing `react-hooks/purity` / `react-hooks/set-state-in-effect` violations from a separate prior bump (`eslint-plugin-react-hooks` 5.2.0 → 7.1.1, #1185). That's a distinct, source-code-level fix and out of scope here to keep this PR focused on the dependency/test-runner breakage.

A changeset is included (`.changeset/fix-client-peer-dependency-mismatches.md`).


tupe12334:
Blocked (not from this PR's diff): CI has 2 failing checks, but both are pre-existing breakage already on \`main\` (same errors on main's latest \`Lint\`/\`client (vitest)\` runs), not introduced by this dependency realignment:
- \`client (typecheck + lint)\`: 8 pre-existing "Cannot call impure function during render" / "Calling setState synchronously within an effect" errors (react-hooks eslint rule) across the client, e.g. \`client/src/shell/CommandPalette.tsx:46\`.
- \`prebuilt-client-html-fresh\`: \`prebuilt.html\` is stale vs. the client bundle.

Since \`main\` itself is currently red on the same checks, this isn't something safe to auto-fix as part of this PR — needs a judgment call on whether to fix upstream first or bypass here.

— PR comment commander

tupe12334:
Blocked: pushed a fix for the CI failures that were in scope for this PR (regenerated `prebuilt-client.html` via `cargo check`, which also resolved `client (vitest)`'s pre-existing `ERR_PACKAGE_PATH_NOT_EXPORTED` failure — confirmed passing now). `client (typecheck + lint)` still fails, but that's the same 8 pre-existing react-hooks errors already broken on `main` (unrelated to this dependency realignment, see #1252).

However, that same regeneration surfaced a new CodeQL alert (high severity, `js/incomplete-url-scheme-check`) in the rebuilt bundle at `prebuilt-client.html:62` — a protocol-allowlist check in vendored react-router-dom code that doesn't include `vbscript:`. It wasn't present before the dependency bump changed the bundled output, and isn't something introduced by application code in this repo. It looks like a false positive against minified third-party library internals rather than an app-level XSS sanitizer, but a high-severity security alert isn't something safe to auto-dismiss — needs a judgment call on whether to dismiss the alert as a false positive or investigate further.

— PR comment commander

tupe12334:
Blocked: superseded by main. `bc9da2e` ("feat(ui): remove the legacy Yew UI, make the React client the default UI", merged 2026-07-19) already fixed this exact peer-dep break on `main`, but via the opposite direction — it bumped `react`/`@types/react` up to `^19.2.7`/`^19.2.3` to match the already-bumped `react-dom@^19.2.7`, and separately pinned `@vitejs/plugin-react` down to `^5.2.0`. This PR instead downgrades `react-dom`/`@types/react-dom` back to `^18.3.1`/`^18.3.7` to match `react@^18.3.1`.

Both resolve the peer-dep mismatch, but they're rival directions on the same dependency set — merging this PR as-is would revert `react-dom` from 19 back to 18 against `react@^19.2.7` already on main, reintroducing a peer-dep mismatch (just inverted) rather than fixing one. Not a mechanical conflict resolution — needs an owner decision on whether anything here (e.g. the changeset wording) is still worth cherry-picking, or whether to just close as superseded.

— PR comment commander

Issue #894, opened 2026-07-02
cargo doc fails on main: broken intra-doc link to private crate::sync::crontab_bin


tupe12334:
## Problem

The \`cargo doc\` CI check currently fails on \`main\` (confirmed at commit \`9a20b6a\`, and reproduced by running \`cargo doc --workspace --no-deps --document-private-items\` locally with \`RUSTDOCFLAGS=-D warnings\`, matching \`.github/workflows/doc.yml\`):

\`\`\`
error: unresolved link to \`crate::sync::crontab_bin\`
   --> src/routines/service.rs:593:8
    |
593 | /// ([\`crate::sync::crontab_bin\`]) closes for crontab I/O (issue #175) — a test that forgets to
    |        ^^^^^^^^^^^^^^^^^^^^^^^^ no item named \`crontab_bin\` in module \`sync\`
    |
error: could not document \`moadim\`
\`\`\`

## Root cause

\`src/routines/service.rs\`'s doc comment on \`sh_bin()\` links to \`[\`crate::sync::crontab_bin\`]\`, but \`crontab_bin()\` in \`src/sync/mod.rs\` is a bare private \`fn\` (no \`pub\`/\`pub(crate)\`), so rustdoc can't resolve the cross-module link even with \`--document-private-items\`.

This is unrelated to any in-flight PR's own diff — it's a standing break on \`main\` that will surface as a \`cargo doc\` failure on every open/new PR's CI run regardless of what that PR touches.

## Fix

Either:
- Change the doc comment in \`src/routines/service.rs:593\` to plain text (drop the \`[\`...\`]\` intra-doc link syntax), or
- Mark \`crontab_bin\` \`pub(crate)\` in \`src/sync/mod.rs\` so the link resolves.

The former is the smaller, lower-risk change since \`crontab_bin\`'s privacy looks intentional.

---
This issue was opened by the Daemon + Landing auto-improve & auto-merge lint/docs PRs routine of moadim. cc @tupe12334

tupe12334:
This appears already fixed on `main`.

PR #897 (merged 2026-07-02, commit 615f50f) applied exactly the fix recommended here: it replaced the `` [`crate::sync::crontab_bin`] `` intra-doc link on `sh_bin()`'s doc comment with plain text, without touching `crontab_bin`'s privacy. That PR's description explicitly says "Filed as issue #894" but its commit/PR body never included a "Closes #894" keyword, so this issue was never auto-closed even though the underlying bug was resolved.

[...]

PR #1252, opened 2026-07-17
fix(client): resolve react-hooks/purity and set-state-in-effect lint errors

tupe12334:
## Why

The `eslint-plugin-react-hooks` bump to 7.1.1 (#1185) enabled stricter React
Compiler lint rules. Applied against pre-existing code, they surface 8 real
errors that currently fail `pnpm --filter client lint` on `main` — part of
both CI's `client-lint` job and the local pre-push hook:

- `react-hooks/purity` (4x): `Date.now()` called directly during render in
  `RefreshControl`, `RoutineFlags`, and `RoutineHistory` (twice). Render is
  supposed to be a pure function of props/state; an impure clock read there
  can produce unstable output across re-renders.
- `react-hooks/set-state-in-effect` (4x): `setState` called synchronously
  inside a `useEffect` body in `RoutinesPage` (the `?history=<id>` deep link,
  and pruning stale selections when the routine list changes), `SettingsPage`
  (seeding the prompt editor's draft from the fetched value), and
  `CommandPalette` (resetting query/selection on open).

This isn't a change any single contributor's PR introduces — it's a standing
break that surfaces on any `client/` touch (or a clean `pnpm --filter client
lint`/pre-push run), the same category of issue as #894 (the now-fixed
`cargo doc` intra-link break). It is unrelated to #1251, which fixes a
separate `@vitejs/plugin-react`/`vite` peer-dependency mismatch that
currently blocks `pnpm --filter client test` from starting at all — that
issue is out of scope here and this PR doesn't touch any dependency version.

## What changed

- Added `client/src/lib/useNow.ts`, a small shared hook that reads the clock
  inside a `setInterval` timer effect (not during render) and returns the
  ticking value. Reused it in `RefreshControl` (replacing its existing
  dummy-tick-counter workaround), `RoutineFlags`, and `RoutineHistory`
  (which previously computed a one-shot, non-ticking `Date.now()` snapshot
  directly in the render body).
- Moved the 4 flagged `setState`-in-effect calls to either a lazy `useState`
  initializer (`RoutinesPage`'s initial page from the `history` search param,
  since it only matters on the first render) or a guarded render-time update
  keyed on the previous prop/value (`RoutinesPage`'s stale-selection prune,
  `SettingsPage`'s draft seeding, `CommandPalette`'s reset-on-open) — the
  pattern React's own docs recommend for "[Adjusting state when a prop
  changes](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)".
  No intended behavior change; `CommandPalette`'s DOM focus side effect stays
  in its own `useEffect` since focusing an input is a legitimate external-system
  side effect, not a `setState` call.

## Checks run locally

- `pnpm --filter client lint` — 0 errors (was 8 errors, 1 pre-existing
  unrelated warning remains)
- `pnpm --filter client typecheck` — clean
- `pnpm --filter client test` — can't run on `main` as-is due to the
  unrelated #1251 peer-dep break (`vitest` fails at config-load with
  `ERR_PACKAGE_PATH_NOT_EXPORTED`), confirmed pre-existing via `git stash`.
  Verified this PR's changes directly by temporarily layering them on top of
  #1251's fix in a local scratch branch (not part of this PR): all 331 tests
  pass, including a new `useNow.test.ts`.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D
  warnings`, `cargo test --workspace`, `cargo llvm-cov --fail-under-lines
  100 --ignore-filename-regex 'src/main\.rs'`, `linecheck --max-lines 500` —
  all green (unaffected, this PR only touches `client/`).
- `typos` on all touched files — clean.

## Changeset

Added `.changeset/fix-client-react-hooks-lint-violations.md` (patch).

### DIFF
--- a/.changeset/fix-client-react-hooks-lint-violations.md
+++ b/.changeset/fix-client-react-hooks-lint-violations.md
@@ -1,0 +1,27 @@
+---
+"moadim": patch
+---
+
+fix(client): resolve `client-lint`'s 8 `react-hooks/purity` and `react-hooks/set-state-in-effect` errors
+
+The `eslint-plugin-react-hooks` bump to 7.1.1 (#1185) enabled stricter React
+Compiler rules that flag pre-existing code: four `Date.now()` calls during
+render (`react-hooks/purity`) and four `setState` calls synchronously inside
+a `useEffect` (`react-hooks/set-state-in-effect`). This left `pnpm --filter
+client lint` — part of both CI's `client-lint` job and the local pre-push
+hook — failing on `main` for any contributor who runs it, independent of
+what their own change touches.
+
+- Added a shared `useNow()` hook (`client/src/lib/useNow.ts`) that reads the
+  clock inside a timer effect instead of during render, and reused it in
+  `RefreshControl`, `RoutineFlags`, and `RoutineHistory` (which previously
+  each read `Date.now()` directly in their render body).
+- Moved four `setState` calls (`RoutinesPage`'s deep-link page and stale-selection
+  prune, `SettingsPage`'s draft-seeding, `CommandPalette`'s reset-on-open) out
+  of `useEffect` and into a lazy `useState` initializer or a guarded
+  render-time update, per React's own "Adjusting state when a prop changes"
+  guidance — no behavior change intended.
+
+No dependency versions changed here; unrelated to #1251, which fixes a
+separate `@vitejs/plugin-react`/`vite` peer-dependency break that currently
+also blocks `pnpm --filter client test` from even starting.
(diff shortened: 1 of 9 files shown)


tupe12334:
Blocked (not from this PR's diff): `client (vitest)` fails with the pre-existing \`@vitejs/plugin-react\`/vite6 peer-dep breakage that's also currently red on \`main\` (\`ERR_PACKAGE_PATH_NOT_EXPORTED\`). Every other check, including \`client (typecheck + lint)\`, passes — this PR's own react-hooks fix is good.

The fix for the peer-dep issue lives in sibling PR #1251, which is itself paused pending an owner call on a new CodeQL alert. Once #1251 is resolved and merged, syncing this branch onto main should clear the remaining failure — nothing to do here until then.

— PR comment commander

PR #1253, opened 2026-07-17
chore(lint): enable clippy::unseparated_literal_suffix workspace-wide

tupe12334:
## Why

`root/Cargo.toml` and `ui/Cargo.toml` each have their own `[lints.clippy]` table (`ui` has no `workspace = true` inheritance, per the many prior lint-parity chores in this repo's history — `format_collect`, `cast_lossless`, `similar_names`, etc.). `clippy::unseparated_literal_suffix` — which denies a numeric literal whose type suffix isn't underscore-separated from its digits (e.g. `500u64` instead of `500_u64`) — had never been enabled in either table, despite the codebase already enforcing the sibling `unreadable_literal` convention (require `_` grouping in *digit* groups) for large integers.

Without the separator, a literal like `100000u64` reads ambiguously at a glance — is that suffix part of the number or not — the same readability problem `unreadable_literal` already guards against for the digit side.

## What

- Added `unseparated_literal_suffix = "deny"` to both `Cargo.toml` and `ui/Cargo.toml`, with the same style of doc comment as every other lint in each table.
- Fixed the 48 violations this surfaced (27 root, 21 `ui`) via `cargo clippy --fix` — every one is a mechanical `Nu8`/`Nu64`/`Nusize`/`Ni32`-style literal (mostly test fixtures, plus one production site in `routines/cleanup/disk_cap.rs`) gaining an underscore before its suffix. No behavior change.

## Checks run locally

- `cargo fmt --check` — clean
- `cargo clippy -p ui --target wasm32-unknown-unknown --all-targets -- -D warnings` — clean (mirrors CI's `clippy-ui-wasm` job)
- `cargo test --workspace` — 1075 + 313 passed
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main.rs'` — 100% line coverage maintained
- `typos`, `linecheck --max-lines 500` — clean

## Changeset

Added `.changeset/enable-clippy-unseparated-literal-suffix.md` (patch).

---


tupe12334:
Blocked (not from this PR's diff): 3 failing checks, none caused by this clippy-only change:
- `client (typecheck + lint)` and `client (vitest)`: the same pre-existing client breakage already tracked on #1251 (vite/plugin-react peer-dep) and #1252 (react-hooks lint) — this PR touches no `client/` files.
- `prebuilt-html-fresh`: rebuilding `ui/` locally (same pinned trunk 0.21.14, `cargo check`) reproduces byte-identical output to what's already committed on this branch — the staleness CI reports isn't reproducible from a contributor machine, pointing to an environment-dependent non-determinism in the wasm build rather than real drift introduced by this PR's diff (a pure `clippy::unseparated_literal_suffix` fix doesn't change compiled output).

Nothing here is safe to auto-fix — needs a judgment call (likely: land #1251/#1252 first, and separately investigate why `prebuilt-html-fresh` disagrees with a local rebuild).

— PR comment commander

PR #1254, opened 2026-07-17
chore(lint): enable clippy::unreachable workspace-wide

tupe12334:
## Why

`clippy::all` (already `deny`-level in both `Cargo.toml` and `ui/Cargo.toml`) doesn't cover the `unreachable!()` macro — it's a `restriction`-group lint, opted into individually like the ~40 other restriction lints this repo already deny-lists (`unwrap_used`, `todo`, `mem_forget`, etc.).

A daemon process (or the running Yew UI) that hits `unreachable!()` panics outright instead of returning a structured error/response, and a match arm that looks provably impossible today can become reachable after an innocuous refactor elsewhere — the panic then only surfaces at runtime, in production, exactly the failure class `unwrap_used`/`todo`/`unimplemented` already guard against in this codebase. This locks in the same guarantee for the last remaining "process-killing macro" left unguarded.

## What

- Added `unreachable = "deny"` to both `Cargo.toml` and `ui/Cargo.toml` (the `ui` crate has its own `[lints.clippy]` table and doesn't inherit root's extended deny-list, so both need their own entry — mirrors the existing lint-parity pattern used for every other restriction lint here).
- Fixed the one violation this surfaced, in `src/build/ui.rs`'s `base64_encode`: a `match chunk.len() { 3 => ..., 2 => ..., 1 => ..., _ => unreachable!() }` where the wildcard arm is genuinely impossible (`bytes.chunks(3)` never yields a chunk outside 1..=3). Rewrote it without a `match`, indexing the guaranteed-present first byte directly and reading the optional second/third bytes via `.get(..).map_or(0, ..)`, so there's no catch-all arm left to guard against.
- `ui/src` had zero violations already.

## Verified locally

- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets --all-features` — clean (0 warnings)
- `cargo test --workspace` — 313 passed, 0 failed
- `cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'` — 100% lines, gate passes
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items` — clean
- Wrote a standalone Rust harness comparing the old `match`-based `base64_encode` against the new `.get()`-based version across chunk lengths 0..20 — byte-for-byte identical output, confirming no behavior change.

Not touching `client/` — no vitest impact.


tupe12334:
Blocked (not from this PR's diff): 3 failing checks, none caused by this clippy-only change:
- `client (typecheck + lint)` and `client (vitest)`: the same pre-existing client breakage already tracked on #1251 (vite/plugin-react peer-dep) and #1252 (react-hooks lint) — this PR touches no `client/` files.
- `prebuilt-html-fresh`: rebuilding `ui/` locally (same pinned trunk 0.21.14, `cargo check`) reproduces byte-identical output to what's already committed on this branch — the staleness CI reports isn't reproducible from a contributor machine, pointing to an environment-dependent non-determinism in the wasm build rather than real drift introduced by this PR's diff (a pure `clippy::unreachable` fix doesn't change compiled output).

Nothing here is safe to auto-fix — needs a judgment call (likely: land #1251/#1252 first, and separately investigate why `prebuilt-html-fresh` disagrees with a local rebuild).

— PR comment commander

PR #1261, opened 2026-07-19
chore(lint): enable clippy::expect_used in the root crate

tupe12334:
## Why

`unwrap_used` is already forbidden in production code (root `Cargo.toml`) because a panic in a long-running daemon kills the server without a graceful shutdown. But `.expect()` panics exactly the same way `.unwrap()` does — it only adds a custom message to the same abrupt failure mode — and `expect_used` was never enabled, so `unwrap_used` alone left an unguarded back door: swap `.unwrap()` for `.expect("message")` and the panic-safety guarantee silently stops applying.

## What

- Enabled `expect_used = "deny"` in `Cargo.toml`'s `[lints.clippy]` (root crate only this round; `ui/` is a separate `[lints]` table and left for a follow-up, matching how `mem_forget` was split into two PRs for root vs `ui`).
- Fixed the 25 violations this surfaced across `src/`:
  - Most (a missing parent directory, a `toml`/`serde_json` re-serialization, a loopback socket address, the process's own executable path, a piped file handle) were converted to proper `Result` propagation — `?`, `ok_or_else`, or a `let else` early return for a `()`-returning best-effort function — matching this crate's existing `unwrap_used`-driven error-handling style.
  - A handful are genuinely provable "can't happen" invariants: an id checked to exist earlier in the same function with the lock held continuously since (`src/routines/service_update.rs`), `Stdio::piped()` set a few lines above the matching `.take()` (`src/sync/mod.rs`), a `?` a line above that already returns on `None` (`src/service/macos.rs`), and the daemon's own executable path being unresolvable in a `String`-returning helper with no `Result` to propagate through (`src/sync/routines.rs`). These are kept as `.expect()` behind a scoped, reasoned `#[allow]` — the same pattern this file already uses for `mem_forget`/`clippy::zombie_processes`.
- `build.rs` (and the `src/build/` modules it pulls in via `#[path = ...]`) carries its own crate-level `#![allow(clippy::expect_used, reason = "...")]`: it's a build script, not the daemon binary, so a `.expect()` panic there just aborts `cargo build` with a message — the intended failure mode, same class of exemption already granted to test code (`allow-expect-in-tests = true` added to `clippy.toml`, mirroring the existing `allow-unwrap-in-tests`).
- Added a changeset (`.changeset/enable-clippy-expect-used.md`).

## Verification

Ran locally, matching CI's exact invocations from `.github/workflows/lint.yml` and `doc.yml`:
- `cargo fmt --check` — clean
- `cargo clippy -p ui --target wasm32-unknown-unknown --all-targets -- -D warnings` — clean (unaffected, `ui` has its own `[lints]` table)
- `cargo test --workspace` — 1075 + 340 passed, 0 failed
- `RUSTDOCFLAGS=-D warnings cargo doc --workspace --no-deps --document-private-items` — clean
- `linecheck --max-lines 500 $(find src ui/src -name '*.rs')` — clean

No behavior change outside the `.expect()` → `Result` conversions themselves (which only change the failure path from "panic the process" to "return an error the caller already handles").

---


tupe12334:
Blocked (not from this PR's diff): 2 failing checks, both pre-existing breakage already on `main` (its latest `Test` workflow run is red on the same jobs), not introduced by this clippy-only change:
- `client (vitest)`: pre-existing `ERR_PACKAGE_PATH_NOT_EXPORTED` (vite/plugin-react peer-dep mismatch), already tracked on #1251 — this PR touches no `client/` files.
- `cargo llvm-cov (100% line floor)`: total coverage is 99.03%, pre-existing gaps in `routines/defaults/mod.rs`, `handlers.rs`, `run_history.rs`, `service_log_tail.rs`, `service/common.rs`, `service/linux.rs`, `sync/mod.rs`, `utils/atomic.rs`, `utils/claude_json.rs` — same gaps already tracked on #1201. A pure `clippy::expect_used` lint change doesn't touch test coverage.

Nothing here is safe to auto-fix — needs #1251 to land (fixes vitest) and the coverage gaps from #1201 closed separately.

— PR comment commander

PR #1265, opened 2026-07-19
chore(lint): enable clippy::redundant_type_annotations workspace-wide

tupe12334:
## Why

Continues the incremental clippy lint-enablement pattern already established in this repo (`unreachable`, `panic`, `expect_used`, `unseparated_literal_suffix`, etc.). `clippy::redundant_type_annotations` rejects a `let` binding whose explicit type annotation exactly matches what the compiler would already infer — the annotation adds nothing beyond what the initializer already states, so it's noise a reader has to cross-check against the inferred type instead of trusting.

## What

- Enabled `redundant_type_annotations = "deny"` in both the root `Cargo.toml` and `ui/Cargo.toml` (the `ui` crate has its own `[lints.clippy]` table and doesn't inherit root's deny-list, per the existing lint-parity pattern used for every prior lint chore).
- No code changes needed — `cargo clippy --workspace --all-targets -- -D warnings` is clean; the workspace (both `src/` and `ui/src`) already had zero violations, so `deny` just locks that in.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --workspace` (340 passed)
- [x] `.changeset/enable-clippy-redundant-type-annotations.md` added (patch bump on `moadim`)


tupe12334:
Blocked: `prebuilt-html-fresh` fails because this PR's `ui/` lint changes altered the trunk-built wasm bundle. I regenerated `prebuilt.html` locally with the exact pinned toolchain (`trunk 0.21.14`, `cargo check`) per the check's own instructions, but the resulting bytes are reproducible on my machine (verified across two runs) yet don't match what the GitHub-hosted runner produces — almost certainly an environment-dependent difference (e.g. absolute paths under `~/.cargo/registry` baked into the wasm) rather than a real UI bug.

I couldn't close that gap: no Docker daemon available here to replicate the `ubuntu-latest` runner byte-for-byte, and the automation token lacks the `workflow` OAuth scope needed to push a temporary artifact-producing workflow to get GitHub Actions to build it directly.

`cargo llvm-cov` and `client (vitest)` failures are unrelated pre-existing breakage already on `main`, not from this PR.

Needs either: a maintainer regenerating `prebuilt.html` from a byte-matching environment, or granting `workflow` scope to this automation's token so it can use CI itself to produce the artifact.

— PR comment commander

PR #1276, opened 2026-07-19
feat(routines): per-routine environment variables via [env] + routine.local.toml

### LINKED ISSUE #408: feat(routines): per-routine environment variables / secrets (routine.local.toml + [env] table) — agents can't get a scoped, per-routine token
> Suggestion filed by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. Unassigned on purpose — self-assign to approve.

## Problem

A routine's agent inherits the daemon operator's **entire** login-shell environment (see `build_routine_command` in `src/routines/command.rs`: the crontab line runs under `/bin/sh -l`, sourcing `~/.profile` so the agent gains `GH_TOKEN`, API keys, etc.). There is no way to set or scope environment **per routine**:

- Every routine on a host shares one global `GH_TOKEN` / credential set. A routine that should act against a different GitHub org (or with a least-privilege, scoped token) can't — they all run as the same identity.
- Non-secret per-routine config (e.g. a model override, a base URL, a feature flag the agent reads) has nowhere to live.
- Cron jobs already have a documented `job.local.toml` for "secrets and machine-specific overrides that stay off-git" (README → Directory layout). **Routines have no equivalent** — an asymmetry between the two scheduled-work primitives.

`Routine` (`src/routines/model.rs`) has no `env` field, and there is no `routine.local.toml`.

## Motivation

- **Least privilege / multi-target:** scope each routine to exactly the token/identity it needs instead of leaking one global credential into every agent session.
- **Parity:** close the gap with cron jobs' `job.local.toml` secret-override story.
- **Off-git secrets:** keep per-routine secrets out of the git-tracked `routine.toml`, mirroring the existing `*.local.toml` convention.

## Proposed approach (sketch — open to discussion)

- Add an optional `[env]` table to `routine.toml` (non-secret defaults, git-tracked) and/or a `routine.local.toml` sibling (untracked, gitignored like `job.local.toml`) for secrets.
- Inject these as `export KEY=<shell-quoted value>` statements in `build_routine_command`, **after** the curated `PATH` export and before launch, so they override profile-inherited values for that session only.
- Reuse the existing `shell_quote` helper; validate keys (`[A-Za-z_][A-Za-z0-9_]*`) and reject newline-injecting values, consistent with the cron-handler validation hardening already tracked elsewhere.
- Surface the keys (names only, never values) in the routine API/UI for visibility.

## Acceptance criteria

- [ ] `routine.toml` accepts an optional `[env]` map (string→string); absent/empty behaves exactly as today.
- [ ] An untracked `routine.local.toml` (gitignored) can supply/override env values; its keys win over `routine.toml` `[env]`.
- [ ] Injected vars are present in the agent's environment at run time and override the profile-inherited values for that run only (verified by a `command`-builder test asserting the emitted `export` statements and precedence).
- [ ] Keys are validated and values are shell-quoted; an invalid key or an injection attempt is rejected on create/update (not written to the crontab).
- [ ] Secret **values** never appear in API responses, the UI, `agent.log`, or `routine.toml`; only key names are exposed.
- [ ] README/Architecture docs document the new `[env]` table and `routine.local.toml`, mirroring the `job.local.toml` section.

## Out of scope

- Encrypted-at-rest secret storage / a secrets backend (keychain, Vault). This issue is the file-based `*.local.toml` parity feature only; a backend can be a later follow-up.

tupe12334:
## Problem

A routine's agent inherits the daemon operator's entire login-shell environment (`build_routine_command` spawns under `sh -lc`, which sources `~/.profile`). There's no way to scope credentials/config per routine — every routine on a host shares one `GH_TOKEN`, and there's no place for non-secret per-routine config either. Cron jobs have `job.local.toml`-style off-git overrides; routines had no equivalent.

## What this does

- **`routine.toml` `[env]` table** (tracked, non-secret): a `string → string` map on `Routine`, persisted/loaded through `RoutineToml`. Absent/empty is a no-op — identical behavior to before this change.
- **`routine.local.toml` sibling** (untracked): a gitignored file next to `routine.toml`, read only at launch time (never held on the in-memory `Routine`/`RoutineStore`, so it can never leak into an API response). Its keys win over `routine.toml`'s `[env]` on conflict. Its `.local.` infix already matches the `*.local.*` pattern every routine's generated `.gitignore` seeds (the same convention as `state.local.toml` / `prompt.compiled.local.md`), so no `.gitignore` changes were needed.
- **Injection**: `build_routine_command` merges both sources (`BTreeMap`, so the emitted statements are deterministic) and emits `export KEY=<shell-quoted value>` right after the curated `PATH` export and before anything else runs — overriding profile-inherited values for that run only. Reuses the existing `shell_quote` helper.
- **Validation**: keys must match `[A-Za-z_][A-Za-z0-9_]*`; values must not contain newlines (shell-injection hardening). Enforced on `svc_create`/`svc_update` (REST + MCP) — an invalid `[env]` entry 400s before anything is persisted or the crontab is touched. `routine.local.toml` is hand-edited and never goes through the API, so the same check runs defensively at launch time: a malformed entry there is dropped (with a `log::warn!`) instead of corrupting the single-line launch command.
- **Redaction**: `Routine::env` is `#[serde(skip_serializing)]` — it never appears in a JSON response, full stop. `RoutineResponse` instead carries a derived `env_keys: Vec<String>` (sorted, deduplicated key names from both sources, values dropped immediately after reading). Verified with a test that serializes a `RoutineResponse` holding a real secret value and asserts the value string never appears in the JSON. The crontab line itself never embeds the launch command (it just invokes `moadim schedule trigger <id>`), and the built command carrying `export` statements is never logged anywhere — confirmed by grepping every `build_routine_command` call site.
- **Defaults routines**: fixed `defaults::reconcile` to preserve a user's `env` across a spec-driven reconcile, the same way `tags`/`model` are preserved (the mechanical struct-literal fixups needed for the new field would otherwise have silently wiped it).
- **Docs**: README's directory-layout diagrams, the routine field table, and a new "Environment variables" section; Architecture.md's tree diagram.

## Testing

- New tests: `src/routines/command_env_tests.rs` (merge/precedence/ordering/validation-at-launch-time), `service_field_validation_tests.rs` (create/update validation + JSON redaction), `routine_storage_tests.rs` (TOML round-trip, `routine.local.toml` reading, gitignore-pattern match).
- `cargo build --workspace --all-targets`, `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo clippy -p ui --target wasm32-unknown-unknown --all-targets -- -D warnings`, `cargo build -p ui --target wasm32-unknown-unknown`, and `cargo test --workspace` (1100 + 340 tests) all pass. `apis/openapi.json` regenerated and committed (confirmed via the `committed_spec_is_current` test).

## Out of scope

- No UI editor for `[env]` was added — the API/MCP already accept it on create/update, and `env_keys` is available for a future UI to render; building that editor felt like a separate follow-up rather than part of closing this issue.
- Encrypted-at-rest secret storage (explicitly called out as out of scope in the issue).

Closes #408.


tupe12334:
Blocked: added the missing changeset (`unreleased-entry` was failing). Three remaining CI failures are genuine code issues from this PR's own diff, not automatable:
- `cargo doc`: 3 broken intra-doc links pointing at stale paths — `crate::routines::command::build_routine_command`, `crate::routines::model::RoutineResponse::from_routine`, and `build_routine_command` (in `src/routine_storage.rs:159`, `:172`, and `src/routines/service_validate.rs:218`) — module no longer has a `command`/`model` submodule at those paths.
- `linecheck`: `src/routines/model.rs` (509 lines) and `src/routine_storage_tests.rs` (573 lines) exceed the 500-line threshold — need splitting.
- `client (typecheck + lint)`: `apis/openapi.json`'s `Routine` schema change (new `env` field) makes several client test fixtures (`heatmapMath.test.ts`, `overviewLogic.test.ts`, `StatsBar.test.tsx`, `filter.test.ts`, `routineState.test.ts`, `commandPaletteMatch.test.ts`, `dayTimelineMath.test.ts`) no longer satisfy the generated `Routine` type — fixtures need the new field(s) added.

All three need judgment calls (doc-link fix, file split, fixture updates) beyond automated CI-fix scope.

— PR comment commander

PR #1348, opened 2026-07-20
test(service): cover svc_update's invalid-env-key reject branch

tupe12334:
## Why

\`svc_update\` calls the same \`validate_env\` helper as \`svc_create\`, but only the
\`svc_create\` path had a test for the invalid-env-key rejection branch
(\`svc_create_rejects_invalid_env_key\`). The \`svc_update\` call site
(\`validate_env(env)?\` in \`service_update.rs\`) was untested, leaving
\`cargo llvm-cov --fail-under-lines 100\` — the repo's own CI gate and pre-push
hook — short by one line.

## What

Adds \`svc_update_rejects_invalid_env_key\`, mirroring the existing
\`svc_update_rejects_blank_repository_url\` test shape: submits an update with an
invalid env key, asserts it's rejected with \`AppError::BadRequest\`, and asserts
the routine's env map is left untouched.

No behavior change — test-only coverage fix. Includes a changeset per
\`.changeset/config.json\`.

## Verification

- \`cargo fmt --check\` — clean
- \`cargo clippy --all-targets -- -D warnings\` — clean
- \`cargo test\` — all pass
- \`cargo llvm-cov --fail-under-lines 100 --ignore-filename-regex 'src/main\.rs'\` — \`service_update.rs\` now at 100% (was 99.52%, missing this one line)


tupe12334:
Blocked: CI failing on this PR's own diff, not automatable —
- `linecheck`: https://github.com/moadim-io/daemon/actions/runs/29786418279/job/88498835765 — `src/routines/service_field_validation_tests.rs` is 510 lines (threshold 500).

Needs a judgment call on how to split the test file (e.g. following the existing pattern of splitting large test modules) — beyond automated CI-fix scope.

— PR comment commander

PR #479, opened 2026-06-20
docs(readme): document the built-in claude agent's python3 + tmux prerequisites (#404)

tupe12334:
## Why

The built-in `claude` agent's `setup` step (in `src/routines/agents/claude_code/setup.rs`) shells out to **`python3`** to pre-seed `~/.claude.json` — the workspace-trust dialog and MCP-server approvals — so the unattended session never blocks on an interactive prompt.

This dependency was **source-only and unchecked**: on a host without `python3`, the setup step fails and the agent run **silently no-ops** while the routine still reports a healthy (green) status. That's a confusing footgun on minimal hosts (CI runners, fresh containers) where `python3` isn't preinstalled — the loop looks fine but never actually fires the agent. Tracked as #404.

## What

Adds a short **"Built-in `claude` agent prerequisites"** note to the README's Routines/Agents section documenting:

- **`python3`** — required by the agent's `setup` step; absence → silent no-op behind a green status.
- **`tmux`** — every routine run is launched inside a tmux session, so the binary must be installed.

## Scope

Docs-only. No behavior change. The changelog gate is exempt (fires only on `src/`/`ui/` changes); `typos` passes clean.

---
This pull request was opened by the **Daemon + Landing auto-improve PRs** routine of moadim.

PR #1362, opened 2026-07-21
fix(routines): surface unresolvable setup-step interpreter as unhealthy

### LINKED ISSUE #404: Default `claude` agent silently requires `python3` (undocumented, unchecked) — a python-less host makes the routine no-op with a green health dot
## Problem

The built-in default **`claude`** agent config (`src/routines/agents/claude_code/setup.rs`) runs its workspace-trust seeding through an inline **`python3 -c '...'`** `setup` step:

```toml
setup = '''python3 -c 'import json,os,sys,tempfile,fcntl; ...' "$WB"'''
```

`python3` is therefore a **hard runtime dependency** of the default `claude` routine, but:

- it is **not documented** anywhere (the README lists no prerequisites beyond Rust/Cargo; only `tmux` is even implied, and that is itself tracked separately in #187), and
- it is **never checked** — there is no PATH/binary probe for it. `grep -rn "python3" src` shows it exists only in this one setup string; there is no `command_exists`/`which` guard for setup-step dependencies.

On a host without `python3` (minimal Linux images, some macOS setups where only `python` exists, stripped containers), the setup step fails. Because the setup step is **not fail-fast** (#287), the agent still launches — but `~/.claude.json` was never seeded, so Claude Code blocks on the "Do you trust this folder?" dialog and the run **silently no-ops** until the max-runtime watchdog reaps it. The routine health dot stays green the whole time.

This is the same class of silent-failure as the unchecked `tmux` dependency (#187) and the unverified agent `command` binary (#383), but neither covers the **`python3` interpreter invoked from inside the default claude `setup` string** — that dependency is invisible to a binary-on-PATH check of the agent `command` (`claude`).

## Motivation

- A fresh `cargo install moadim` + default `claude` routine on a python-less host fails with **zero diagnostic signal** — green health, empty `agent.log`, reaped after an hour.
- The dependency is undiscoverable: nothing in the docs or health output mentions it.
- Hardening here closes the last leg of the "unchecked hard dependency" trio (tmux #187, agent command #383, and now the setup-step interpreter).

## Acceptance criteria

- [ ] Document the `python3` requirement of the default `claude` agent (README prerequisites and/or the `claude.toml` comment), alongside the existing `tmux` note.
- [ ] Detect a missing `python3` and surface it, rather than failing silently. At minimum one of:
  - [ ] routine health reports unhealthy (or a warning) when a routine's resolved agent has a `setup` step whose first token isn't on `PATH`, **or**
  - [ ] `moadim doctor` (#323) probes for `python3` and reports it among its silent-failure checks.
- [ ] Consider de-risking the dependency itself (optional / separate follow-up): replace the inline `python3` seeding with a shell/jq implementation or a small native step, so the default agent has no interpreter dependency. Document the chosen direction.
- [ ] Tests cover the "setup-step interpreter missing" path (e.g. the health/doctor probe flags it), reusing the existing PATH-probe seam where available.

---
_This issue was filed as a non-binding suggestion by the **Suggest new GitHub issues (daemon + landing)** routine of moadim. Self-assign to approve it. cc @tupe12334_

tupe12334:
## Summary

Closes #404's remaining behavioral gap.

- PR #479 already documented the built-in `claude` agent's `python3`/`tmux` prerequisites in the README ("Docs-only. No behavior change.").
- PR #902 already surfaced a missing `python3` at the **daemon** level: a startup warning and `GET /api/v1/health`'s `dependencies.python3` flag. Its own changelog explicitly noted the gap this PR closes: *"though the affected routine's own health dot still shows green."*

What was still missing: the **per-routine** health status. `RoutineResponse.agent_registered` only checks that `~/.config/moadim/agents/<agent>.toml` exists and parses — it says nothing about whether the agent's `setup` step can actually run. A routine using the built-in `claude` agent on a python-less host had a perfectly valid, registered config, so it kept reading "healthy" even though the launch command's existing fail-fast guard (`{ setup; } || { ...; exit 1; }` in `build_routine_command`) was guaranteed to abort the run in `setup` before the agent ever launched.

- Add `RoutineResponse.agent_setup_available: bool` (`src/routines/model.rs`), mirroring the existing `agent_command_available` field: `true` unless the agent config has a `setup` step whose first whitespace-delimited token (the interpreter/binary it shells out to, e.g. `python3`) fails to resolve on the daemon's `PATH`.
- New `setup_step_interpreter`/`setup_step_available` helpers in `src/routines/command_path_resolution.rs`, following the exact seam `agent_command_available`/`tmux_available` already established for this kind of PATH probe.
- Wire the new field into the client's `routineHealth()` (`client/src/pages/routines/filter.ts`): a routine whose `setup` step can't run now falls into the same "AGENT MISSING" badge `agent_registered` already uses, instead of "healthy".
- `apis/openapi.json` / `client/src/api/schema.gen.ts` regenerated to match (the latter is gitignored/build-generated, not committed).
- `prebuilt.html` regenerated (CI's `prebuilt.yml` enforces this stays in lockstep with `client/`).
- README's "Built-in `claude` agent prerequisites" section updated to describe the now-closed gap.
- Changeset added.

## Test plan

- [x] `cargo test --workspace` — 1154 passed (new Rust unit tests: `setup_step_interpreter`/`setup_step_available` in `command_bin_resolution_tests.rs`, `RoutineResponse::from_routine` coverage in `model_tests.rs`)
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `client`: `pnpm run typecheck`, `pnpm run lint`, `pnpm test` — all clean (364 tests passed, including a new `routineHealth` case for the unresolvable-`setup`-interpreter path)
- [x] `make readme-check` — clean


tupe12334:
Blocked: CI failing on this PR's own diff, not automatable —
- `cargo doc`: https://github.com/moadim-io/daemon/actions/runs/29829834244/job/88631634205 — broken intra-doc link in `src/routines/command_path_resolution.rs:62`: `[`super::command::build_routine_command`]` resolves to no item named `command` in module `command` (wrong path).

Needs the doc link path corrected against the actual item location — a small judgment call on the right target, beyond automated CI-fix scope.

— PR comment commander

PR #1402, opened 2026-07-23
Enable clippy::struct_excessive_bools lint

### LINKED ISSUE #1401: Enable clippy::struct_excessive_bools lint
## What it catches

`clippy::struct_excessive_bools` (pedantic) fires on any struct with more than 3 `bool` fields.

## Why it improves code quality

Past that count, a call site constructing or reading the struct positionally (or a `Default`-then-mutate sequence) can no longer tell which `true`/`false` means what without cross-checking field names against their declarations. Worse, `n` independent bools imply `2^n` possible states, most of which were never meant to be reachable together — a state machine or a two-variant enum usually expresses the real, restricted set of valid combinations far more precisely than a pile of loose flags.

This repo already denies a large set of pedantic/nursery clippy lints in `[lints.clippy]` (`Cargo.toml`) as part of an ongoing effort to keep the codebase's error-handling, panic-safety, and API-shape conventions machine-enforced rather than convention-only. This lint is a natural addition to that set.

## Scope

`cargo clippy --all-targets -W clippy::struct_excessive_bools` currently surfaces exactly one violation: `routines::model::RoutineResponse`, a `#[serde(flatten)]` HTTP response DTO with four bool fields (`agent_registered`, `agent_command_available`, `agent_setup_available`, `is_running`). Those bools are independent, already-documented probes, not combinatorial state, and collapsing them into an enum would break the JSON shape existing API clients parse — so that struct is the right candidate for a scoped, reasoned `#[allow]`, matching the pattern this file already uses for `mem_forget`/`clippy::zombie_processes`.

---
Opened by the moadim routine **"Clippy lint improvements → issue + PR (Rust repos) → Slack"**.

tupe12334:
## Summary
- Enables `clippy::struct_excessive_bools` (pedantic) in `[lints.clippy]`, denying structs with more than 3 `bool` fields — past that count a caller can't tell which flag means what without cross-checking field names, and the `2^n` combinations invite unreachable-in-practice states.
- Fixes the one violation this surfaced: `routines::model::RoutineResponse` (a `#[serde(flatten)]` HTTP response DTO) carries four independent, already-documented probe bools (`agent_registered`, `agent_command_available`, `agent_setup_available`, `is_running`). These aren't combinatorial state, and collapsing them into an enum would change the JSON shape existing API clients parse — so it's exempted with a scoped, reasoned `#[allow]`, matching the `mem_forget`/`clippy::zombie_processes` pattern already used in `Cargo.toml`.

Closes #1401

## Test plan
- [x] `cargo clippy --all-targets` — clean
- [x] `cargo build` — passes
- [x] `cargo test` — 1164 passed, 0 failed

---
Opened by the moadim routine **"Clippy lint improvements → issue + PR (Rust repos) → Slack"**.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -510,3 +510,15 @@ equatable_if_let = "deny"
 # unlike the stable one, sorts in place without allocating a scratch buffer. The codebase is
 # already clean, so `deny` locks that in.
 stable_sort_primitive = "deny"
+# Reject a struct with more than 3 `bool` fields — past that count, a call site constructing or
+# reading the struct positionally (or a `Default`-then-mutate sequence) can no longer tell which
+# `true`/`false` means what without cross-checking field names, and the 2^n combinations invite
+# states that were never meant to be reachable together. The one violation this surfaced,
+# `routines::model::RoutineResponse`, is a `#[serde(flatten)]` HTTP response DTO whose four bools
+# (`agent_registered`, `agent_command_available`, `agent_setup_available`, `is_running`) are
+# independent, already-documented probes with their own doc comments, not combinatorial state —
+# collapsing them into an enum would also break the JSON shape existing API clients parse. Kept as
+# a scoped, reasoned `#[allow]` on that struct, the same pattern already used for
+# `mem_forget`/`clippy::zombie_processes` above. The rest of the codebase is already clean, so
+# `deny` locks that in.
+struct_excessive_bools = "deny"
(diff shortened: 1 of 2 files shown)


tupe12334:
Closing as superseded: `origin/main` already contains `struct_excessive_bools = "deny"` in `Cargo.toml` and the scoped `RoutineResponse` allow with reason, so the PR intent is already landed.

PR #1417, opened 2026-07-27
feat(routines): pre-clone/cache declared repositories into the workbench (#466)

### LINKED ISSUE #466: Routines declare structured `repositories` (url + branch) but the daemon only prints them — pre-clone/cache declared repos into the workbench instead of re-cloning inside every agent run
## Problem

A routine's `repositories` field is **structured** — each entry has a `repository` URL and an optional `branch` (`src/routines/model.rs`), and these are validated on create/update (`validate_repositories` in `src/routines/service.rs`). The daemon already knows, declaratively, exactly which repos (and branches) a routine needs.

Yet at run time the daemon does **nothing** with that knowledge except render it as prose. `compose_prompt` (`src/routines/command.rs`) emits:

```
# Workbench
You are working in an empty directory. These repositories are relevant — clone any you need:
- <url> (branch <branch>)
```

Each run gets a **fresh, empty workbench** and the burden of actually cloning is pushed entirely onto the agent, on every fire. Consequences:

- **Re-clone every run.** A `*/5` routine against a large repo re-clones from scratch ~288×/day — network, time, and disk churn that the per-run workbench TTL then reaps, only to redo it next fire. There is no shared cache / reference mirror.
- **Clone failures are a silent no-op.** This repo's recurring failure mode (cf. #404, #383, #443, #445) repeats here: a bad URL, an unreachable host, a missing-credentials clone, or a non-existent `branch` fails *inside* the agent session. The run still counts as "fired" (green dot), and the declared `branch` is never actually checked out — nothing on the daemon side ever verifies the declared repos are reachable.
- **The `branch` field is advisory only.** It's printed into the preamble but never enforced; an agent may ignore it and clone the default branch, so the routine silently operates on the wrong ref.

## Motivation

The daemon owns the workbench lifecycle and already has the validated repo/branch list. Using it to materialize the working copy (instead of only describing it) makes runs faster, cheaper, and fail *loudly and early* — turning a class of silent no-ops into a recorded run error before the agent even starts.

## Proposed direction (sketch, open to discussion)

- Add an opt-in step that, before launching the agent, clones each declared `repository` (at the declared `branch`) into the workbench — e.g. `<workbench>/<repo-name>/`.
- Back it with a persistent bare-mirror cache under the config dir (e.g. `~/.config/moadim/cache/<host>/<owner>/<repo>.git`) and clone with `--reference` / a local `git clone <mirror>` so repeated runs reuse fetched objects instead of re-downloading. (Interacts with the workbench disk-cap discussion in #398.)
- Surface a clone/fetch failure as a **run-level error** (ties into #453 "runs record no exit status") rather than letting it disappear inside the agent.
- Keep it backward compatible: when the step is disabled (default off, or no repositories declared), behaviour is exactly today's "empty workbench + clone-it-yourself" preamble.

## Acceptance criteria

- [ ] A configurable mechanism (per-routine or global flag) makes the daemon pre-clone each declared `repository` at its `branch` into the run workbench before the agent launches.
- [ ] Repeated runs of the same routine reuse a local object cache/mirror instead of a full re-clone each fire (verified by a test or a documented benchmark).
- [ ] A failed clone/fetch (bad URL, unreachable host, missing branch) aborts the run with a recorded, surfaced error instead of a silent in-agent failure.
- [ ] When the feature is disabled or no repositories are declared, the current empty-workbench + prose-preamble behaviour is unchanged.
- [ ] Behaviour and config keys are documented in the README / Architecture.md; the cache directory honours the existing config-path conventions.

---

_This issue was suggested by the "Suggest new GitHub issues (daemon + landing)" routine of moadim — it's an unassigned suggestion. Approve it by self-assigning; otherwise it can be closed. cc @tupe12334_

tupe12334:
## Problem

A routine's `repositories` field (`{ repository, branch }`) was only rendered into the prompt preamble as prose — "these repositories are relevant — clone any you need" — and the daemon never acted on it. Every fire pushed the clone onto the agent, with no shared cache and no daemon-side verification:

- A `*/5` routine against a large repo re-cloned from scratch ~288x/day.
- A bad URL, unreachable host, or a `branch` that doesn't exist upstream failed silently *inside* the agent session while the run still counted as "fired".
- The `branch` field was advisory only — never enforced.

## Fix

`build_routine_command` now pre-clones each declared repository into the workbench before the agent launches (`routines::command_repositories::clone_repository_stmts`):

- Each repo is backed by a persistent local mirror under `~/.config/moadim/cache/`, keyed by its exact URL. The first run of a given URL does `git clone --mirror`; every later run — of any routine referencing that same URL — only `git fetch`es it, reusing already-downloaded objects instead of re-cloning from the remote.
- The workbench copy is a fast, local `git clone` from that mirror (no network), checked out at the declared `branch` when set.
- A failure at either step (bad URL, unreachable host, missing branch) aborts the run with the reason recorded in `agent.log`, mirroring the existing `cp prompt.md`/`setup`/`tmux` guards, instead of failing silently inside the agent or not at all.
- A routine with no declared `repositories` is completely unaffected — no clone/fetch statements are emitted (verified by a test).
- `compose_prompt`'s preamble now points the agent at where each repo already lives (`./<name>`) instead of asking it to clone.

Kept intentionally minimal: no new per-routine config flag — declaring `repositories` is itself the opt-in signal, so no schema/API surface changed. `README.md`/`Architecture.md` updated to match the new behavior; `apis/openapi.json` and `schemas/routine.schema.json` regenerated for the updated field description.

## Verification

- `cargo test` — 1179 passed, 0 failed.
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` — clean.
- `linecheck --max-lines 500` over `src/**/*.rs` — clean.
- New unit tests cover `repo_dir_name` (URL → directory-name sanitization) and `clone_repository_stmts` (cache-vs-fetch branching, branch checkout, loud-abort-on-failure), plus `build_routine_command` integration tests for the with/without-repositories cases.

Closes #466

---
Opened automatically by the moadim routine: Open issues → fix PRs (owned repos + my orgs).

### DIFF
--- a/.changeset/precache-routine-repos.md
+++ b/.changeset/precache-routine-repos.md
@@ -1,0 +1,5 @@
+---
+"moadim": patch
+---
+
+feat(routines): pre-clone/cache declared repositories into the workbench before the agent launches (#466), instead of only printing them as a "clone any you need" prompt preamble. Each declared repository is backed by a persistent local mirror under `~/.config/moadim/cache/`, keyed by its URL — the first run clones the mirror, every later run of any routine referencing the same URL only `git fetch`es it, so repeated fires reuse already-downloaded objects instead of re-cloning from the remote. A clone/fetch failure (bad URL, unreachable host, a `branch` that doesn't exist upstream) now aborts the run with the reason recorded in `agent.log`, the same as the existing prompt-copy/setup/tmux guards, instead of failing silently inside the agent session.
--- a/Architecture.md
+++ b/Architecture.md
@@ -179,12 +179,16 @@ That crontab line does **not** launch the agent by itself — it invokes the `mo
 the daemon to be running (it is installed as an OS service — launchd/systemd user — for exactly
 this reason). That handler (`routines::service_trigger::svc_trigger_scheduled`) and the manual
 `POST /routines/{id}/trigger` handler both funnel into `spawn_routine_command`, which builds the
-launch script (`routines::command::build_routine_command`) and spawns it in-process: it makes a fresh empty
-workbench under `~/.moadim/workbenches/`, launches the agent **interactively** (no `-p`) in a
-detached tmux session rooted there, and captures output via `pipe-pane`. The prompt reaches the
-agent as a process **argument** (the `{prompt}` placeholder expands to `"$(cat prompt.md)"`), so
-there is no keystroke-injection readiness race. The agent decides whether to clone the listed
-repositories.
+launch script (`routines::command::build_routine_command`) and spawns it in-process: it makes a fresh
+workbench under `~/.moadim/workbenches/`, pre-clones each declared repository into it
+(`routines::command_repositories::clone_repository_stmts`, #466) from a persistent local mirror
+under `~/.config/moadim/cache/` — the first run of a given repo URL clones the mirror, every later
+run (of any routine referencing the same URL) only fetches it — then launches the agent
+**interactively** (no `-p`) in a detached tmux session rooted there, and captures output via
+`pipe-pane`. The prompt reaches the agent as a process **argument** (the `{prompt}` placeholder
+expands to `"$(cat prompt.md)"`), so there is no keystroke-injection readiness race. A clone/fetch
+failure (bad URL, unreachable host, missing `branch`) aborts the run and is recorded in `agent.log`,
+the same as a failed prompt copy or `setup` step.
 
 The spawned script runs under a *login* shell (`sh -lc`) so the user's `~/.profile` is sourced and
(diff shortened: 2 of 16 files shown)


tupe12334:
**PR comment commander**: this PR is otherwise mergeable but blocked by a failing required check:

- `cargo llvm-cov (100% line floor)` — overall coverage is 99.45% (line), below the 100% floor. `src/routines/model.rs` (touched by this PR) is at 99.15% coverage.

This needs a human judgment call on what test(s) to add, so I'm leaving it for you rather than guessing. Will pick this back up automatically once it's green.

PR #1564, opened 2026-08-27
fix(macos): schedule routines in daemon

tupe12334:
Replace macOS routine crontab writes with a daemon-owned 15-second scheduler that uses existing scheduled-trigger dedupe/locks.


tupe12334:
Superseded: replacing the custom scheduler loop with a maintained scheduler library per direction.

PR #1572, opened 2026-08-27
fix(scheduler): persist scheduled fire evidence before launch

tupe12334:
## Summary
- persist one timestamp in each routine scheduled.log before the detached scheduled launcher is attempted
- emit an explicit accepted-fire daemon log with routine id, timestamp, and evidence path
- remove the launcher-side duplicate stamp so each accepted in-process fire has one durable record

## Verification
- cargo test (1,329 tests)
- cargo clippy --all-targets -- -D warnings
- cargo fmt --all -- --check
- cargo llvm-cov --fail-under-lines 100 (repository gate)
- .githooks/pre-push: Rust checks, coverage, linecheck, and client typecheck/lint/tests (536 tests)

The first full hook run exposed a base-suite race in sync::routines::routines_sync_status_tests::successful_routine_sync_clears_previous_crontab_sync_error; the same test passed from a clean origin/main worktree and the complete hook rerun passed.

### DIFF
--- a/.changeset/durable-scheduled-fire-evidence.md
+++ b/.changeset/durable-scheduled-fire-evidence.md
@@ -1,0 +1,5 @@
+---
+"moadim": patch
+---
+
+Persist explicit scheduled-fire evidence before in-process scheduler launch attempts.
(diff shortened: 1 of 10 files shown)


tupe12334:
Closing: this was created outside the requested scope. The requested lesson updates belong in Hermes skills PRs.

